PR #5, opened 2020-05-26
Add build+test CI, make public code platform-independent (ish)

crsaracco:


crsaracco:
All of the `#[cfg]` spam can be cleaned up when we unify the API between the three platforms.

The most common way I've seen to do this is the "platform" pattern: [rtb-rs](https://github.com/rust-dsp/rtb-rs/blob/master/src/platform/mod.rs), [druid](https://github.com/xi-editor/druid/blob/master/druid-shell/src/platform/mod.rs).

MirkoCovizzi:
I'm not a fan of having an example with conditional compilation inside of it. Can we move this one layer down? Also, conditional compilation in the `lib.rs` for platform specific visibility should be handled in a more organized way I think (but to make it compile I independently had to make the same changes as you did pretty much).

We could learn from winit:
https://github.com/rust-windowing/winit/tree/master/src/platform_impl

I feel like the sooner we deal with conditional compilation for platform specificity (in a clean way), the better it's going to be down the road.

crsaracco:
I'm not a fan either, but I figured we'd get to it when we unified the API. My main goal here was to spin up some CI so that we don't accidentally/unwittingly break it in the future :)

PR #28, opened 2020-06-30
Add Timer and manage some Events

MirkoCovizzi:


MirkoCovizzi:
This implements #27 and #22

PR #32, opened 2020-09-02
Add beginnings of abstract Message struct and message receiver. Add compiler argument for choosing between OpenGL and wgpu.

BillyDM:
I made some tweaks to begin integrating `baseview` into `iced_baseview`.

- Added abstract `Message` structs.
- Added `Receiver` trait.
- Added compiler arguments for choosing between OpenGL and wgpu. (I haven't added any wgpu stuff yet).
- Refactored some of the code in `x11/window.rs` into `x11/opengl_util.rs`.
- Made `x11/window.rs` return `Message` messages to the `Receiver`.
- Updated example.

These changes are only tested in Linux.


glowcoil:
Some thoughts:

- We will probably want a better encoding for both keycodes and character inputs.
- We may want to avoid making the platform-specific code generic over the receiver, to avoid forcing users to recompile all of it when they modify their receiver. This might not end up being a big deal at all.

But I don't think either of those things needs to block merging this. Looks good to me!

src/macos/window.rs
```diff
@@ -42,7 +44,13 @@ impl Window {
             current_app.activateWithOptions_(NSApplicationActivateIgnoringOtherApps);
             app.run();
 
-            Window
+            receiver.on_message(Message::Opened(crate::message::WindowInfo {
+                width: options.width as u32,
+                height: options.height as u32,
+                dpi: None,
+            }));
```
crsaracco, line 51: Event handling will need to be in a new thread eventually, FYI.

BillyDM:
Good idea on not having generics for platform-specific code. I'll find a way to do that.

BillyDM:
There might be a crate that could help us with character and key encoding.

src/macos/window.rs
BillyDM: What exactly is the thread lifecycle of a plugin? Or at-least the lowest common demoninator?

My assumption is like this:
1. Host requests to open window by sending a window handle to the GUI object, which constructs itself.
2. On a separate thread, the host sends input events to the GUI object.
3. On a separate thread, the host requests the GUI struct to render its content to the window.
4. Host requests to close the window. The GUI object is destroyed.

crsaracco: It's been a while since I did this last, but [here](https://github.com/crsaracco/vst2-gui-prototypes/blob/master/linux-opengl-vst/src/editor/window/mod.rs)'s the prototype I've made in the past for a VST with pure X11 windowing + OpenGL.

Essentially:
 1. Host calls vst `open()`, giving you a parent window handle. You spin up a new thread, create the window with that parent, and start looping to handle events.
 2. The window you created handles the events itself (in that new thread). There's a "host callback" for plugin -> host communication, if needed.
 3. "request draw" is just another event (`xcb::EXPOSE`), so you handle rendering there (in that new thread). I forget how you're supposed to keep drawing, maybe send yourself an `xcb::EXPOSE` event in a loop? @wrl ?
 4. Host requests to close the window. You have to safely exit the other thread (probably by breaking out of the loop somehow)

wrl: > Event handling will need to be in a new thread eventually, FYI.
Can't be on macOS or Windows, will need to be on Linux.

crsaracco: Oh derp, this was the macos module. X.x

wrl: >     1. Host calls vst `open()`, giving you a parent window handle. You spin up a new thread, create the window with that parent, and start looping to handle events.
> 
>     2. The window you created handles the events itself (in that new thread). There's a "host callback" for plugin -> host communication, if needed.

The host callback will need to make its way over. Hosts handle the possibility of having it called from multiple threads.

>     3. "request draw" is just another event (`xcb::EXPOSE`), so you handle rendering there (in that new thread). I forget how you're supposed to keep drawing, maybe send yourself an `xcb::EXPOSE` event in a loop (in yet another thread)? @wrl ?

I think it'll make more sense to have `draw()` be a method on the handler/responder trait. druid-shell has all of the event types broken out as separate methods, but since baseview is designed as an intermediate layer, having a dedicated event struct makes it easier to just implement `From<baseview::Message> for SpecificToolkitMessage` or similar.

>     4. Host requests to close the window. You have to safely exit the other thread (probably by breaking out of the loop somehow)

I think that, after the client code (plugin, in this case) opens a window, the result of the open call is some sort of channel for sending messages to the window, and then our Message type to has an application-specific variant, something like:

```rust
pub enum Message<AppSpecific> {
    ...
    ApplicationSpecific(AppSpecific)
}
```

When the plugin receives the editor close message, it'll send across a message.indicating that the window should close, which will be handled in the `on_message()` method.

wrl:
I feel like some of the naming could be better (I'm not personally sold on `Receiver`) but not only do I have any concrete suggestions at the moment but I also think we can deal with that incrementally post-merge.

After investigating wgpu, I am of the mind that we should standardise on wgpu to start and not have an option for OpenGL. By the time anybody ships an actual plugin based on this stack, Vulkan (which hit 1.0 in early 2016) will be more than half a decade old. Keeps baseview leaner and smaller as well, which I'm very here for.

Would you mind terribly removing the GL code and replacing it with wgpu in this PR?

BillyDM:
Okay. I'll get to it tomorrow!

wrl:
Thanks for the PR! Really good to get more folks involved.

BillyDM:
* Removed context-creation code.
* Implemented a `Message` channel with `std::sync::mpsc`.
* `Window::open()` now spawns the event loop in a separate thread for x11.
* Added `Window::raw_window_handle()` to get the `RawWindowHandle` for user context creation.

examples/open_window.rs
```diff
@@ -6,5 +10,62 @@ fn main() {
...
+                println!("Window unfocused");
+            }
+            Message::Opened(window_info) => {
+                println!("Window opened, {:?}", window_info);
+            }
+            Message::WillClose => {
+                println!("Window will close");
+            }
+        }
+    }
```
wrl, line 70: This should go back to being a `Receiver` or similar trait. Having all window messages dispatched and handled asynchronously through an MPSC channel creates enormous problems, largely pertaining to de-synchronisation during window system events.

For example, handling resizes asynchronously is a primary reason why resize is laggy in many toolkits.

wrl (changes requested):

wrl:
> I'm unsure how to make the `Receiver` trait not be generic over platform-specific code. Also, does anyone have a better name for what we can call the trait?

Why does `Receiver` have to be generic? Is it because `Message` also needs to be generic?

I'm fine with you calling it `Receiver` for the time being. Down the line I think we'll want to restructure what exactly a `Window` is and we can sort out trait/structure naming at that juncture.

BillyDM:
Because `Receiver` is your application itself. It is how your application receives its messages while not owning its own threads.

BillyDM:
I think `Receiver` should just be called `Application`, that is what `iced` does.

wrl:
The issue with "Application" is that there can be multiple independent baseview windows in the same application (in the plugin case), so I think it's a misnomer. That said, it's fine for now – all struct/trait names are subject to change as we refine and mature the API.

examples/open_window.rs
```diff
@@ -6,5 +10,67 @@ fn main() {
...
+    let (_app_message_tx, app_message_rx) = mpsc::channel::<()>();
+
+    // Send _app_message_tx to a separate thread, then send messages to the GUI thread.
+
+    let _ = baseview::Window::open(window_open_options, my_program, app_message_rx);
+}
+struct MyProgram {}
+
+impl baseview::AppWindow for MyProgram {
+    type AppMessage = ();
```
wrl, line 24: i like having this be an associated type :+1:

wrl (changes requested):

examples/open_window.rs
BillyDM: That is how `iced` does it!

BillyDM:
I also changed the MouseClickType to click_count. I'm not sure why it didn't show up in the commits.

BillyDM:
I added a small wrapper for `RawWindowHandle` because `iced` expects a struct that implements `HasRawWindowHandle`.

wrl (approved):
@BillyDM thanks for doing this work, looks good to me! :+1:

BillyDM:
You're welcome!

src/x11/window.rs
```diff
@@ -34,33 +35,6 @@ impl Window {
...
-                glx::GLX_DOUBLEBUFFER,  1,
-                0
-            ],
-        );
-
-        // The GLX framebuffer config holds an XVisualInfo, which we'll need for other X operations.
-        let x_visual_info: *const xlib::XVisualInfo =
-            unsafe { glx::glXGetVisualFromFBConfig(xcb_connection.conn.get_raw_dpy(), fb_config) };
-
         // Load up DRI2 extensions.
```
crsaracco, line 38: IIRC, I was going to use DRI2 for OpenGL stuff (?). I think we can remove it since we're removing OpenGL.

src/x11/window.rs
```diff
@@ -85,65 +64,41 @@ impl Window {
...
-            // Connection
             &xcb_connection.conn,
-            // Depth
-            unsafe { *x_visual_info }.depth as u8,
-            // Window ID
+            xcb::COPY_FROM_PARENT as u8,
             window_id,
-            // Parent ID
             parent_id,
-            // x
```
crsaracco, line 125: please keep comments at least for the ones that don't have descriptive variable names. You can do something like `/* x= */ 0` or something to minimize the lines on screen if desired.

BillyDM:
Using `RawWindowHandle::Xcb` was causing `wgpu` to panic. Changing it to a `RawWindowHandle::Xlib` appears to have fixed it.

src/x11/window.rs
```diff
@@ -85,65 +64,41 @@ impl Window {
...
-            // y
             0,
-            // width
             options.width as u16,
-            // height
             options.height as u16,
-            // border width
-            0,
-            // class
+            10,
```
crsaracco, line 87: border width 10?

src/x11/window.rs
```diff
@@ -157,36 +112,6 @@ impl Window {
...
-        unsafe {
-            gl::load_with(|n| opengl_util::load_gl_func(&n));
-        }
-
-        // Check GL3 support
-        if !gl::GenVertexArrays::is_loaded() {
-            panic!("no GL3 support available!");
-        }
-
         // TODO: This requires a global, which is a no. Figure out if there's a better way to do it.
```
crsaracco, line 115: I believe this was also part of OpenGL setup, so we can probably remove this.

Hopefully there's no hidden global state in `wgpu-rs` or whatever...?

src/x11/window.rs
```diff
@@ -198,58 +123,59 @@ impl Window {
...
-            panic!("Obtained indirect rendering context");
-        }
-
-        // Display the window
-        xcb::map_window(&xcb_connection.conn, window_id);
-        xcb_connection.conn.flush();
+        // What does this do?
         unsafe {
             xlib::XSync(xcb_connection.conn.get_raw_dpy(), xlib::False);
         }
```
crsaracco, line 129: I think you can replace this with a `xcb_connection.conn.flush();`. I'd suggest moving the one from right after `xcb::map_window` down here.

src/x11/window.rs
```diff
@@ -198,58 +123,59 @@ impl Window {
...
+        let scaling = {
+            let maybe_scaling =
+                get_scaling_xft(&xcb_connection).or(get_scaling_screen_dimensions(&xcb_connection));
+
+            if let Some(scaling) = maybe_scaling {
+                scaling
+            } else {
+                1.0
+            }
+        };
```
crsaracco, line 150:
```rust
let scaling = get_scaling_xft(&xcb_connection)
    .or(get_scaling_screen_dimensions(&xcb_connection))
    .or(Some(1.0))
    .unwrap();
```

src/x11/window.rs
```diff
@@ -198,58 +123,59 @@ impl Window {
...
+        x11_window.run_event_loop();
+
+        x11_window
     }
 
     // Event loop
-    fn handle_events(&self, window_id: u32, ctx: *mut x11::glx::__GLXcontextRec) {
-        let raw_display = self.xcb_connection.conn.get_raw_dpy();
+    fn run_event_loop(&mut self) {
+        //let raw_display = self.xcb_connection.conn.get_raw_dpy();
```
crsaracco, line 177: you can just remove this.

crsaracco (approved):
LGTM, nice work on the architecture. Just some quick nitpicks.

src/x11/window.rs
wrl:
```rust
let scaling = get_scaling_xft(&xcb_connection)
    .or(get_scaling_screen_dimensions(&xcb_connection))
    .unwrap_or(1.0);
```

;)

src/x11/window.rs
BillyDM: Whoops, that was from a copy-and-paste. :p

src/x11/window.rs
crsaracco: haha, nice. that one.

BillyDM: That is nicer! I'll have to remember that syntax more.

src/x11/window.rs
```diff
@@ -138,17 +107,10 @@ impl<A: AppWindow> Window<A> {
...
-                scaling
-            } else {
-                1.0
-            }
-        };
-        println!("Scale factor: {:?}", scaling);
+        let scaling = get_scaling_xft(&xcb_connection)
+            .or(get_scaling_screen_dimensions(&xcb_connection))
+            .or(Some(1.0))
+            .unwrap();
```
wrl, line 113: not to be too much of a jerk about this but i really would prefer an `unwrap_or(1.0)` here instead of a raw `unwrap()` (more idiomatic, also raw `unwrap()` is a bit of a code smell, especially in libraries)

wrl (changes requested):

wrl (approved):

wrl:
:tada:

PR #36, opened 2020-09-06
separate Window::open() and Window::run() on macos

glowcoil:


glowcoil:
Closing this because the API as-is is slightly more future-proof.

PR #40, opened 2020-09-10
Add and refactor events for iced integration

BillyDM:
I set up the framework of events we should target on each platform if possible. I mostly followed the way `iced` structures its events.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -1,5 +1,46 @@
+use crate::KeyCode;
+use std::path::PathBuf;
+
+/// The current state of the keyboard modifiers.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct ModifiersState {
+    pub shift: bool,
+    pub control: bool,
+    pub alt: bool,
+    pub logo: bool,
+}
+
+impl ModifiersState {
+    /// Returns true if the current [`ModifiersState`] has at least the same
+    /// modifiers enabled as the given value, and false otherwise.
+    ///
+    /// [`ModifiersState`]: struct.ModifiersState.html
+    pub fn matches(&self, modifiers: ModifiersState) -> bool {
+        let shift = !modifiers.shift || self.shift;
+        let control = !modifiers.control || self.control;
+        let alt = !modifiers.alt || self.alt;
+        let logo = !modifiers.logo || self.logo;
+
+        shift && control && alt && logo
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum KeyboardEvent {
+    KeyPressed {
+        key_code: KeyCode,
+        modifiers: ModifiersState,
+    },
+    KeyReleased {
+        key_code: KeyCode,
+        modifiers: ModifiersState,
+    },
+    CharacterInput(char),
+    ModifiersChanged(ModifiersState),
+}
+
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
-pub enum MouseButtonID {
+pub enum MouseButton {
     Left,
     Middle,
     Right,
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -31,16 +35,31 @@ pub struct WindowOpenOptions<'a> {
...
+    pub frame_rate: f64,
 }
 
 pub trait AppWindow {
     type AppMessage;
 
     fn build(window_handle: RawWindow, window_info: &WindowInfo) -> Self;
 
-    fn draw(&mut self);
+    fn draw(&mut self, mouse_cursor: &mut MouseCursor);
```
wrl, line 48: why is passing the mouse cursor necessary here?

src/macos/window.rs
```diff
@@ -59,7 +60,7 @@ impl<A: AppWindow> Window<A> {
             let window_info = WindowInfo {
                 width: options.width as u32,
                 height: options.height as u32,
-                scale: 1.0,
+                scale_factor: 1.0,
```
wrl, line 63: `scale` is fine.

wrl (changes requested):

src/macos/window.rs
BillyDM: Both iced an winit calls it `scale_factor`, but it's fine if you just want it to be `scale`.

src/lib.rs
BillyDM: There needs to be a way to set the mouse cursor icon. In `iced` it is set after rendering, so I thought that we could implement it like that as well. Maybe it should be `fn draw(&mut self) -> Option<MouseCursor>` instead?

src/macos/window.rs
wrl: I don't think that the `_factor` adds any additional clarity or information and would indeed prefer we keep it `scale`.

src/lib.rs
wrl: We're going to need more platform operations than just "set mouse cursor," so I don't think this is the right place to do it.

Also, I don't really think that `draw()` is the right place to be doing any of these sorts of operations anyway. They should happen in response to an event, so we'll need to pass some sort of a handle or struct into `on_event()` with operations like "open URL", "set cursor", "warp pointer", "open file selection dialog", etc.

I'm open to reading anything that shows why `iced` handles that in rendering but I don't think it's the right place to handle it.

BillyDM: Do you think the return should be added to `on_event()` instead? We also need a way to tell `baseview` that the user wants to close the window. I'm thinking one of these methods should return a message struct.

wrl: Returning operations means that only one event can happen in response, which isn't always a safe assumption. We need some object which is passed into `on_event()` which can have these operations called on it.

BillyDM: Yeah, I figured. I'm working on a solution now!

BillyDM:
How do you feel about this design?

BillyDM:
Oh, hold on. There is one more thing I want to change.

BillyDM:
Okay, what do you think of this architecture?

src/event.rs
```diff
@@ -1,5 +1,46 @@
...
+    /// [`ModifiersState`]: struct.ModifiersState.html
+    pub fn matches(&self, modifiers: ModifiersState) -> bool {
+        let shift = !modifiers.shift || self.shift;
+        let control = !modifiers.control || self.control;
+        let alt = !modifiers.alt || self.alt;
+        let logo = !modifiers.logo || self.logo;
+
+        shift && control && alt && logo
+    }
+}
```
wrl, line 26: Is there a reason this isn't just covered by the `Eq` auto-derive?

src/event.rs
```diff
@@ -1,5 +1,46 @@
...
+    KeyPressed {
+        key_code: KeyCode,
+        modifiers: ModifiersState,
+    },
+    KeyReleased {
+        key_code: KeyCode,
+        modifiers: ModifiersState,
+    },
+    CharacterInput(char),
+    ModifiersChanged(ModifiersState),
```
wrl, line 39: Will the window receive both a `ModifiersChanged` *and* the respective `KeyPressed`/`KeyReleased` events when a modifier is pressed/released?

src/event.rs
```diff
@@ -8,39 +49,108 @@ pub enum MouseButtonID {
...
+    ///
+    /// When the user drops multiple files at once, this event will be emitted
+    /// for each file separately.
+    FileDropped(PathBuf),
+
+    /// A file was hovered, but has exited the window.
+    ///
+    /// There will be a single `FilesHoveredLeft` event triggered even if
+    /// multiple files were hovered.
+    FilesHoveredLeft,
```
wrl, line 145: Any reason for the semantics here surrounding multiple files? Is this just how the iced abstractions work?

src/event.rs
```diff
@@ -8,39 +49,108 @@ pub enum MouseButtonID {
...
-    WindowResized(WindowInfo), // new (width, height)
-    WindowFocus,
-    WindowUnfocus,
-    WillClose,
+    Interval(f64), // delta time passed
+    Mouse(MouseEvent),
+    Keyboard(KeyboardEvent),
+    Window(WindowEvent),
+    FileDrop(FileDropEvent),
+    Clipboard(Option<String>),
```
wrl, line 155: Let's not have event types here that we don't have implementations for already.

In particular, clipboard events are a bit of a minefield across platforms, and most platforms have more complex data models than simply strings (i.e. differing mime-types).

BillyDM:
Do you like it this way, or do you think we should push requests onto a queue instead?

src/window_state.rs
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+
+    pub fn request_redraw(&mut self) {
+        self.redraw_requested = true;
+    }
+
+    pub fn request_close(&mut self) {
+        self.close_requested = true;
+    }
```
wrl, line 92: No, unfortunately this is not going to work. We *need* to be able to make *synchronous* window system requests from inside the event handler.

When events/requests to/from the window system become buffered and asynchronous, that's when the frames of lag start piling up. We *cannot* buffer our most basic abstractions.

wrl (changes requested):
I don't think that having a separate `WindowState` struct is going to solve the issue here.

What we need is to pass the `Window<A>` into the event handler the way that `WindowState` is being passed in now, and then the `Window<A>` needs `pub` functions for synchronously submitting these requests to the window system.

Whether that happens directly in the `impl<A> Window<A>` blocks and we have our types concrete or whether we have a platform-neutral `Window`/`WindowSystem`/`WindowHandle` type is something that @glowcoil and I disagree on (with @glowcoil being in the "concrete `impl` block" camp and me being in the "platform-neutral trait" camp).

src/event.rs
BillyDM: Oh, this is copy pasted from iced.

src/event.rs
BillyDM: Yeah, again copied from iced.

src/window_state.rs
BillyDM: Okay, I see.

src/event.rs
wrl: No worries! I don't intrinsically have a problem with splitting multiple files into independent events, just wanted to know where the semantics came from.

My preference would be to structure the events close to what the platforms themselves expose here, which, since it would entail a bottom-up design in the platform code, means that maybe let's hold off on the file events for now.

src/event.rs
BillyDM: Oh yeah. Again iced is doing this.

src/window_state.rs
BillyDM: But if we poll for requests immediately after calling `self.app_window.on_event()`, then wouldn't that still be synchronous with the event handler?

wrl: The issue is not necessarily in the *setter* operations that are currently implemented as long as those setters don't have to be called in a particular order, but for *getter* operations (such as "current clipboard contents" or "mouse double click interval", among others) it doesn't work at all unless we copy everything into this `WindowState` before dispatching events, which I don't think makes sense.

wrl: > But if we poll for requests immediately after calling `self.app_window.on_event()` in the same function, then wouldn't that still be synchronous with the event handler?

For single setter calls, yes. Not always a safe assumption to make.

BillyDM: Oh, yeah I see the issue now.

glowcoil:
> Whether that happens directly in the `impl<A> Window<A>` blocks and we have our types concrete or whether we have a platform-neutral `Window`/`WindowSystem`/`WindowHandle` type is something that @glowcoil and I disagree on (with @glowcoil being in the "concrete `impl` block" camp and me being in the "platform-neutral trait" camp).

TBH I'm open to either of these! I don't really have a strong preference.

BillyDM:
Okay, I'll work on a solution tomorrow!

PR #43, opened 2020-09-11
Add and refactor events

BillyDM:
I made a new pull request that only modifies the event structure.

### DIFF
--- a/examples/open_window.rs
+++ b/examples/open_window.rs
@@ -24,42 +24,11 @@ impl WindowHandler for MyProgram {
 
     fn on_event(&mut self, window: &mut Window, event: Event) {
         match event {
-            Event::CursorMotion(x, y) => {
-                println!("Cursor moved, x: {}, y: {}", x, y);
-            }
-            Event::MouseDown(button_id) => {
-                println!("Mouse down, button id: {:?}", button_id);
-            }
-            Event::MouseUp(button_id) => {
-                println!("Mouse up, button id: {:?}", button_id);
-            }
-            Event::MouseScroll(mouse_scroll) => {
-                println!("Mouse scroll, {:?}", mouse_scroll);
-            }
-            Event::MouseClick(mouse_click) => {
-                println!("Mouse click, {:?}", mouse_click);
-            }
-            Event::KeyDown(keycode) => {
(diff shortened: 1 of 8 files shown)


examples/open_window.rs
```diff
@@ -24,42 +24,11 @@ impl WindowHandler for MyProgram {
...
-            Event::WindowFocus => {
-                println!("Window focused");
-            }
-            Event::WindowUnfocus => {
-                println!("Window unfocused");
-            }
-            Event::WillClose => {
-                println!("Window will close");
-            }
+            Event::Interval(delta_time) => println!("Update interval, delta time: {}", delta_time),
```
wrl, line 27: What is the `Interval` event?

BillyDM: Its when the user requests to have a message called in periodic intervals. For example, I may want to process the DSP of my visualizer 30 frames a second or so to save performance, then request a redraw. The plugin will still run at the native frame-rate.

BillyDM: If you have something else in mind for this, I'll remove it.

wrl:
Also, please rebase on top of master.

examples/open_window.rs
wrl: Please remove it for now. While we will need this in the future for timers/animations or whatever, I'm not certain that this is the best way to handle it.

BillyDM:
Also, I am getting this error:
```error[E0658]: use of unstable library feature 'option_zip'
  --> src/x11/window.rs:96:14
   |
96 |             .zip(xcb_connection.atoms.wm_delete_window)
   |              ^^^
   |
   = note: see issue #70086 <https://github.com/rust-lang/rust/issues/70086> for more information```

wrl:
> Also, I am getting this error:
> 
> ```
> error[E0658]: use of unstable library feature 'option_zip'
>   --> src/x11/window.rs:96:14
>    |
> 96 |             .zip(xcb_connection.atoms.wm_delete_window)
>    |              ^^^
>    |
>    = note: see issue #70086 <https://github.com/rust-lang/rust/issues/70086> for more information```
> ```

`Option::zip()` is stable as of 1.46.0. You may need to update your toolchain.

BillyDM:
Oh right

PR #44, opened 2020-09-11
Nonblocking Event Loop, separate WindowHandle.app_run_blocking() call

wrl:
This PR replaces the previous semantics of `Window::open()` blocking indefinitely with `Window::open()` returning a `WindowHandle` immediately. `WindowHandle` then has an `app_run_blocking()` method which replicates the previous semantics.

In standalone applications, `WindowHandle.app_run_blocking()` must be called from `fn main()` in order to keep the app from immediately closing. In plugins, `app_run_blocking()` should *not* be called.

### DIFF
--- a/examples/open_window.rs
+++ b/examples/open_window.rs
@@ -1,16 +1,5 @@
 use baseview::{Event, Window, WindowHandler};
 
-fn main() {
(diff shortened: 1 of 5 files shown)


wrl:
cc @glowcoil @BillyDM

PR #45, opened 2020-09-11
Detect window resize in x11

BillyDM:
I also ran `cargo fmt`


BillyDM:
We may want to add an action that checks formatting.

PR #46, opened 2020-09-13
Add set_mouse_cursor method to X11 Window

BillyDM:
I ended up having to do what `winit` does to accomplish this.


wrl (changes requested):
Could you put the x11 cursor work into a new `src/x11/cursor.rs` file?

wrl (approved):

PR #47, opened 2020-09-15
Add frame_interval_secs to WindowOpenOptions

BillyDM:
To achieve lower input latency in `iced`, I need the frame interval to run at 120fps. Iced sets the `wgpu` swapchain to `Mailbox` mode, so rendering still happens at 60fps. However, this causes less misses for the current frame reducing the perceived latency.

### DIFF
--- a/examples/open_window.rs
+++ b/examples/open_window.rs
@@ -28,6 +28,7 @@ fn main() {
         width: 512,
         height: 512,
         parent: baseview::Parent::None,
+        ..Default::default()
     };
 
     let handle = Window::open::<MyProgram>(window_open_options);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -33,10 +33,24 @@ unsafe impl Send for Parent {}
 pub struct WindowOpenOptions {
     pub title: String,
 
-    pub width: usize,
-    pub height: usize,
+    pub width: u32,
+    pub height: u32,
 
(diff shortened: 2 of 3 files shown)


BillyDM:
I also find that having the default as exactly 60fps feels smoother with less chugging.

wrl:
It seems to me like the issue here is that Iced doesn't provide a mechanism for configuring the wgpu swapchain mode, and so in `Mailbox` mode the draw loop proceeds entirely unsynchronised relative to the vblank.

I'm going to reject this PR because I consider frame timing a window system issue and not one of the application itself. For example, Wayland surfaces receive frame/present requests for drawing their frame, so a Wayland baseview backend wouldn't have a timer at all. Likewise on macOS, we'll be investigating using `CVDisplayLink` instead of a timer.

X11 is a bit of an outlier in that it doesn't provide a reasonable mechanism for vblank notification for the purposes of timing or triggering frame drawing, and that's largely why we have a very simple timer for the time being. If your wgpu swapchain could be created in the `Fifo` mode, you would at the very least synchronise your draw loop to vblank (since presenting the drawn buffer blocks until vblank).

Still, this is something I am very interested in solving in the future, but since frame timing is dictated by the window system, you should consider the X11 frame timer a stopgap hack, and for that reason I'm unwilling to expose any part of it in the public API.

PR #48, opened 2020-10-15
Add UI scaling support

BillyDM:


BillyDM:
Oh yeah, I should fix the ubuntu and mac builds

BillyDM:
Although I notice the size hints only work in the baseview example. The windows doesn't resize as a plugin in Bitwig.

BillyDM:
I don't have a Mac to test it on, and for some reason the data I put into `WindowState<H>` on Windows is getting corrupted for some reason. Do you have any clues?

Please let me know how you feel about this PR. Do you think `baseview` should handle the logic of logical and physical sizes and coordinates, or should the user be the one to handle it? If not, I can change it.

src/lib.rs
```diff
@@ -30,15 +35,18 @@ pub enum Parent {
...
-    pub height: usize,
+pub struct WindowHandle {
+    thread: std::thread::JoinHandle<()>,
+}
 
-    pub parent: Parent,
+impl WindowHandle {
+    pub fn app_run_blocking(self) {
+        let _ = self.thread.join();
+    }
```
wrl, line 45: `WindowHandle` doesn't have the same structure on Windows and macOS, so for the time being it needs to be kept in the platform directories. `app_run_blocking()` in particular is extremely platform-specific.

Please back this change out.

src/window_open_options.rs
```diff
@@ -0,0 +1,74 @@
...
+        initial_size: PhySize,
+        /// The minimum physical width and height
+        min_size: PhySize,
+        /// The maximum physical width and height
+        max_size: PhySize,
+        /// Whether to keep the aspect ratio when resizing (true), or not (false)
+        keep_aspect: bool,
+    },
+}
+
```
wrl, line 33: I would really prefer it if you split out the window size policy min/max work into a separate PR, so that this PR could just be about handling scaling and the logical/physical translation.

wrl (changes requested):
Also merged https://github.com/RustAudio/baseview/pull/49 which changes the `WindowHandler` API slightly and has some conflicts with your PR. Nothing major – should be easy to merge or rebase against master.

wrl:
macOS is going to be really interesting. We may in the future have to redo the scaling policies, but, hey, we'll cross that bridge when we come to it.

BillyDM:
I'm just trying to get it to build for now.

wrl:
Yeah absolutely. The thing on Mac is that all of the sizes and coordinates are *already* in logical units at the window system level, and the physical pixels come into play at a different layer. We'll just have to address the sizing policy slightly differently.

wrl:
Some notes from Discord:

> so, i'm going to merge this, but i have some reservations about the amount of combinations that are possible here
> and i would like you to follow up on this PR by writing some documentation about what each of the scaling policies do at window open when combined with both logical and physical sizes
> because to be frank i'm struggling to figure out all of the possible outputs
> and relatedly i'm struggling to figure out how i would want to construct window open options
> like if i pass in the physical size of the window, does that imply WindowScalePolicy::NoScaling?
> should that turn off backing-store scaling on mac?

Going to merge, but want to get this sorted out ASAP. There's no real sense in having a comprehensive API if people

 1. Aren't going to use much of the scope of it
 1. Struggle to accomplish the simple/common case with it.

PR #49, opened 2020-10-20
replace `WindowHandler::build()` with closure passed to `Window::open()`

wrl:
This PR makes it substantially easier to construct a window which incorporates state from outside the window context.


wrl:
approved by @glowcoil on discord

PR #50, opened 2020-10-21
Clean up dpi features

BillyDM:


src/window_open_options.rs
```diff
@@ -1,46 +1,24 @@
...
 }
 
 /// The options for opening a new window
 #[derive(Debug)]
 pub struct WindowOpenOptions {
     pub title: String,
 
-    /// The size information about the window
-    pub size: WindowSize,
+    /// The logical size of the window
```
wrl, line 17: imo add "These dimensions will be scaled by the scaling policy specified in `scale`." to this comment

wrl (changes requested):

src/window_open_options.rs
BillyDM: How about "These dimensions will be scaled by the scaling policy specified in scale. Mouse coordinates will be translated back to the logical size."

wrl (approved):

Issue #51, opened 2020-10-24
macOS: support events


greatest-ape:
I thought it would make sense to add an issue for macOS event support, so that there is a location for a structured discussion of how to implement it. There is an [analogous implementation in antonok's vst_window crate](https://github.com/antonok-edm/vst_window/blob/master/src/platform/macos/event_source.rs), which could be used as an inspiration.

In my understanding, the NsView created in Window::open needs to be subclassed with an event handling delegate, which stores a pointer to the WindowHandler trait implementation (?) so that its methods call be called in the event handling delegate's _extern "C"_-declared methods. Such an implementation would differ from the one in _vst_window_ in that it would handle events synchronously instead of sending them over a channel that is polled over elsewhere.

PR #52, opened 2020-10-25
macOS: basic event handling

greatest-ape:
Adds barebones handling of mouse down events on macOS (#51) with a lot of code from https://github.com/antonok-edm/vst_window. The open_window example works.

Possible problems that I can think of:

- No reference to the created Window is stored in the WindowHandle, which means it can't be accessed later and cleanup must be handled by macOS (?).
- Parent window functionality has not been tested
- AsIfParented support is not implemented

If this looks like something we could make use of, I can have a look at implementing other events as well.

### DIFF
--- a/src/macos/window.rs
+++ b/src/macos/window.rs
@@ -10,7 +10,7 @@ use cocoa::appkit::{
     NSApplicationActivationPolicyRegular, NSBackingStoreBuffered,
     NSRunningApplication, NSWindow, NSWindowStyleMask,
 };
-use cocoa::base::{id, nil, NO, YES};
+use cocoa::base::{id, nil, NO};
 use cocoa::foundation::{NSAutoreleasePool, NSPoint, NSRect, NSSize, NSString};
 
 use objc::{msg_send, runtime::Object, sel, sel_impl};
@@ -62,12 +62,9 @@ impl Window {
         let mut window = match options.parent {
             Parent::WithParent(parent) => {
                 if let RawWindowHandle::MacOS(handle) = parent {
-                    let ns_window = handle.ns_window as *mut objc::runtime::Object;
                     let ns_view = handle.ns_view as *mut objc::runtime::Object;
 
                     unsafe {
-                        ns_window.setAcceptsMouseMovedEvents_(YES);
-
                         let subview = create_view::<H>(&options);
 
                         let _: id = msg_send![ns_view, addSubview: subview];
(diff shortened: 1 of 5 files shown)


greatest-ape:
Several problems have been identified and fixed. Some remaining work to be done is:

- [ ] Get CursorMoved events when user is not dragging
- [ ] Test if parented views work
- [ ] Test if AsIfParented views work (AUv2 only)
- [ ] Accept keyboard events. Does acceptsFirstResponder or something similar need to be setup to accomplish this?
- [ ] Does the NSApp need to be created **before** setting up the non-parented window and get passed into the WindowHandle for _app_run_blocking_?

src/macos/window.rs
```diff
@@ -38,56 +57,145 @@ impl Window {
...
-            let rect = NSRect::new(
-                NSPoint::new(0.0, 0.0),
-                NSSize::new(
-                    window_info.logical_size().width as f64,
-                    window_info.logical_size().height as f64
-                ),
+            (*window_state.window.ns_view).set_ivar(
+                WINDOW_STATE_IVAR_NAME,
+                Arc::into_raw(window_state.clone()) as *mut c_void
             );
```
wrl, line 149: this `.clone()` here seems funky – do we need that?

wrl (changes requested):

src/macos/window.rs
```diff
@@ -38,62 +57,164 @@ impl Window {
...
+    pub(super) fn trigger_event(&mut self, event: Event){
+        self.window_handler.on_event(&mut self.window, event);
+    }
 }
 
+
 unsafe impl HasRawWindowHandle for Window {
     fn raw_window_handle(&self) -> RawWindowHandle {
+        let ns_window = self.ns_window.unwrap_or_else(
+            ::std::ptr::null_mut
```
glowcoil, line 206: just as a style nit, I would prefer `.unwrap_or(::std::ptr::null_mut())`

src/macos/window.rs
```diff
@@ -38,62 +57,164 @@ impl Window {
...
 
-            WindowHandle
+impl <H: WindowHandler>WindowState<H> {
+    /// Returns a mutable reference to an WindowState from an Objective-C callback.
+    ///
+    /// `clippy` has issues with this function signature, making the valid point that this could
+    /// create multiple mutable references to the `WindowState`. However, in practice macOS
+    /// blocks for the entire duration of each event callback, so this should be fine.
+    #[allow(clippy::mut_from_ref)]
+    pub(super) fn from_field(obj: &Object) -> &mut Self {
```
glowcoil, line 190: This function should be marked `unsafe`, since it is possible to create undefined behavior by calling it from purely safe code. The comment should just describe the requirements for calling it safely (not using it to create two simultaneous `&mut`s to the same WindowState).

greatest-ape: Fixed

src/macos/window.rs
greatest-ape: Fixed

glowcoil (approved):
Looks good to me! I tested and it works fine in both the parented and unparented cases.

Issue #53, opened 2020-10-28
macOS: keyboard event handling


greatest-ape:
We should do this similarily to how it is done in druid.

Cocoa documentation is [here](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/HandlingKeyEvents/HandlingKeyEvents.html#//apple_ref/doc/uid/10000060i-CH7-SW1). [xcb info about keycodes here](https://xcb.freedesktop.org/XCBToDoKeyboard/), [iced keyboard event](https://docs.rs/iced_native/0.2.2/iced_native/input/keyboard/enum.Event.html), [druid implementation](https://github.com/linebender/druid/blob/master/druid-shell/src/platform/mac/keyboard.rs), [useful crate](https://crates.io/crates/keyboard-types)

Snippet to extract string from key event:

```rust
let characters = unsafe {
    let ns_string = NSEvent::characters(event);

    let start = NSString::UTF8String(ns_string);
    let len = NSString::len(ns_string);

    let slice = ::std::slice::from_raw_parts(
        start as *const u8,
        len
    );
                  
    ::std::str::from_utf8_unchecked(slice)
};
```

PR #56, opened 2020-11-13
Add support for macOS and X11 key events with code from druid

greatest-ape:
#53


src/event.rs
```diff
@@ -1,12 +1,12 @@
 use crate::{WindowInfo, Point};
 
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum KeyboardEvent {
```
glowcoil, line 4: We could further simplify by replacing this struct entirely with `keyboard_events::KeyboardEvent`, since it already the `key_state` field for pressed vs. released, right?

greatest-ape: Yeah, and later, we can add composition events: https://docs.rs/keyboard-types/0.5.0/keyboard_types/struct.CompositionEvent.html

greatest-ape: Fixed

greatest-ape:
Status report:

- [x] Keyboard events on macOS in parented mode
- [x] Keyboard events on macOS in non-parented mode
- [ ] Keyboard events on X11 in parented mode
- [x] Keyboard events on X11 in non-parented mode, not locale-sensitive

PR #57, opened 2020-11-14
Add windows keyboard event support with druid code

greatest-ape:


greatest-ape:
Localised keyboard events work in parented and non-parented mode. There is a Windows chime when pressing keys in parented mode in REAPER for some reason.

glowcoil:
Looks good! I was unable to get any key events in the parented case in Ableton on Windows, but it worked in Reaper, so we can look into a solution for that later.

PR #60, opened 2020-11-24
Ignore unknown X11 events

WeirdConstructor:
I am getting many (many) `Unhandled event type: 65` prints when testing my GUI.
That event number is way beyond the defined X11 events, and I guess it makes
sense to just not print anything for these events for now.

### DIFF
--- a/src/x11/window.rs
+++ b/src/x11/window.rs
@@ -411,6 +411,10 @@ impl Window {
                 );
             }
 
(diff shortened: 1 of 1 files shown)


glowcoil:
IMO we should just not print anything at all for unhandled events.

WeirdConstructor:
Yes, that would be the best option. Didn't know how long the debug prints were going to  be needed.

glowcoil:
Removing the print in #63.

PR #79, opened 2020-12-17
macOS: WindowEvent::WillClose

greatest-ape:


greatest-ape:
So the clean up code doesn’t actually run. The frame timer has a reference to the view, and at least with iced_baseview, one more reference is kept to it at this point (meaning that in that particular case, retain count is 3 when cleaning should happen).

So the first step towards a solution is to use a different timer which keeps no reference:

https://developer.apple.com/documentation/corefoundation/1543570-cfrunlooptimercreate
https://docs.rs/core-foundation/0.9.1/core_foundation/runloop/type.CFRunLoopTimerCallBack.html

We control flow also needs to be done in a different manner, since apparently, the host sends the close signal to the vst plugin, which should tell baseview to close the window. If it frees its own references first,  the reference count to check for will be 1.

greatest-ape:
Closing in favor of #84

PR #83, opened 2020-12-20
Add ReceivedCharacter event and add unicode input for linux

BillyDM:


BillyDM:
Okay, this isn't actually working at the moment. I tried setting my keyboard to spanish mode, and none of the special characters are being recognized. I guess I need to try another approach.

PR #87, opened 2021-01-09
tweak x11 frame interval

BillyDM:
After tweaking the frame interval in X11, I found 14ms feels smoother than using 15ms, especially when using OpenGL.


glowcoil:
I don't think this is a good idea. The shorter we make the frame interval, the more we draw twice per vblank and waste one of the frames. The right way to make things smoother would be to hook vsync.

PR #93, opened 2021-03-06
switch from xcb crate to xcb-sys

glowcoil:


glowcoil:
Closing this as there are better options now. The latest version of the `xcb` crate seems to have largely been rewritten and doesn't appear to have the same soundness issues anymore, and if we're updating the X11 bindings then `x11rb` would also be a great option.

PR #99, opened 2021-07-17
Fix clippy warnings

yamadapc:
Sorry for a very uninteresting PR.


robbert-vdh:
Hi! I'm going through some of the old PRs now. This PR has been ignored for a while, and in the two years since creating it the emitted Clippy warnings have changed. There are a couple new ones, and from the ones addressed by this PR only `mut_from_ref` is still relevant. I think we could tackle that one differently by changing the `*mut WindowState` into a `*mut RefCell<WindowState>`. Then `WindowState::from_field` would return a `&RefCell<WindowState>` instead of `&mut Self. That would not only get rid of the Clippy lint, but would also prevent unsoundness if reentrant function calls do happen.

Since this PR not really applicable anymore (sorry for that!) I'll close it for now. I'll create a replacement PR that fixes the current Clippy lints.

PR #103, opened 2021-11-03
add ability to close window from user code, add HostWindowHandle

BillyDM:
I've added a method that allows the user to close the window from their code. This is fairly straightforward.

I've also added two new structs: `HostHandle` and `HostWindowHandle`. These are used when opening a window in parented mode. The `HostWindowHandle` is sent back to the user (which they will use as the UI handle in baseplug), and the `HostHandle` is stored internally in the baseview window.

When the `HostWindowHandle` is dropped, then the window is closed automatically Also vice versa, if the `HostHandle` drops unexpectedly, it will trip a flag in the `HostWindowHandle`, alerting the host/parent (I'm not sure how useful the latter case is, but it was trivial to add in).

I've tested the parentless version of this on Linux, Windows, and Mac. Also the parentless mode in Mac OS was broken anyway, so I've also made changes to fix that as well.

I have also tested this in a plugin context in Bitwig under Linux and Windows. This seems to fix an issue I was having before with the window not opening again after being closed.

Although I have *not* tested the plugin context in Mac yet since I can't seem to get baseplug working for some reason.


BillyDM:
Alright, I've merged the latest commit to my branch. I tested the MacOS code on both unparented and as a plugin, and they both seem to work now (except for manually closing a plugin window from the user's code since I'm not sure how to get that to work. Custom "close window" buttons aren't common in plugins anyway so it's probably not a big deal right now).

I've also renamed the `HostWindowHandle` struct to `ChildWindowHandle` (and `HostHandle` to `ParentHandle`).

src/window.rs
```diff
@@ -12,6 +12,8 @@ use crate::win as platform;
 #[cfg(target_os = "linux")]
 use crate::x11 as platform;
 
+pub use platform::ChildWindowHandle;
```
glowcoil, line 15: Could we rename this to just `WindowHandle`?

src/window.rs
```diff
@@ -55,6 +61,10 @@ impl<'a> Window<'a> {
     {
         platform::Window::open_blocking::<H, B>(options, build)
     }
+
+    pub fn request_close(&mut self) {
```
glowcoil, line 65: Can we just call this `close()`, rather than `request_close()`? "Request close" makes more sense to me as referring to what happens when the user clicks the X on a window, which just sends an event to the application requesting that the window be closed, whereas when the `WindowHandler` calls this method on the `Window`, the window should just be closed.

src/win/window.rs
```diff
@@ -283,19 +341,24 @@ unsafe fn unregister_wnd_class(wnd_class: ATOM) {
 struct WindowState {
     window_class: ATOM,
     window_info: WindowInfo,
+    parent_handle: Option<ParentHandle>,
     keyboard_state: KeyboardState,
     mouse_button_counter: usize,
     handler: Box<dyn WindowHandler>,
     scale_policy: WindowScalePolicy,
     dw_style: u32,
+    destroy_msg_id: u32,
```
glowcoil, line 350: This can just be a const at the top of the file; I don't think we need to keep it as a member of `WindowState`.

src/win/window.rs
```diff
@@ -51,6 +54,52 @@ unsafe fn generate_guid() -> String {
...
+        if let Some(hwnd) = self.hwnd {
+            unsafe {
+                PostMessageW(hwnd, self.destroy_msg_id, 0, 0);
+            }
+        }
+    }
+}
+
+struct ParentHandle {
+    child_window_dropped: Arc<AtomicBool>,
```
glowcoil, line 80: This shouldn't need to be an `Arc<AtomicBool>` on Windows, just an `Rc<Cell<bool>>`, since everything on Windows happens in a single thread.

src/win/window.rs
```diff
@@ -247,7 +297,15 @@ unsafe extern "system" fn wnd_proc(
                 unregister_wnd_class(window_state.borrow().window_class);
                 SetWindowLongPtrW(hwnd, GWLP_USERDATA, 0);
             }
-            _ => {}
+            _ => {
+                if msg == destroy_msg_id {
+                    // FIXME: handler should decide whether window stays open or not
```
glowcoil, line 302: Whether the window is closed via `Window::close()` or by dropping the `ChildWindowHandle`, the handler doesn't need to be able to decide whether the window stays open or not, since they're already the one who decided to close it in the first place. The handler should get a choice in the case where the X button is clicked (`WM_CLOSE` event on Windows), which is a different thing, and explicitly has the semantics of "requesting the window to be closed but allowing the application to decide".

src/win/window.rs
```diff
@@ -51,6 +54,52 @@ unsafe fn generate_guid() -> String {
...
+    child_window_dropped: Arc<AtomicBool>,
+}
+
+impl ChildWindowHandle {
+    pub fn window_was_dropped(&self) -> bool {
+        self.child_window_dropped.load(Ordering::Relaxed)
+    }
+}
+
+impl Drop for ChildWindowHandle {
```
glowcoil, line 69: I guess this is a matter of preference, but I personally would like it better if there were an explicit `close()` method on the `ChildWindowHandle`, and dropping the handle didn't do anything. This sort of mirrors the semantics of `JoinHandle` in `std`, where there's an explicit `join()` method, and dropping the handle just detaches the thread.

src/win/window.rs
```diff
@@ -51,6 +54,52 @@ unsafe fn generate_guid() -> String {
...
+        }
+    }
+}
+
+struct ParentHandle {
+    child_window_dropped: Arc<AtomicBool>,
+}
+
+impl ParentHandle {
+    pub fn new() -> (Self, ChildWindowHandle) {
```
glowcoil, line 84: Could we get rid of the `ParentHandle` struct and just store the `Rc<Cell<bool>>` `window_closed` flag directly in `WindowState` (and then set it to true in `WM_NCDESTROY`), and then also just have `Window::open` directly return a `ChildWindowHandle` (which `open_blocking` can just ignore)? That will let us get rid of the awkward dance where the `HWND` is an option that starts out empty and then later gets set by `Window::open`.

glowcoil (changes requested):

src/window.rs
BillyDM: Possibly. I just don't want it to cause confusion with the `RawWindowHandle` crate.

src/window.rs
```diff
@@ -38,7 +42,9 @@ impl<'a> Window<'a> {
         platform::Window::open_parented::<P, H, B>(parent, options, build)
     }
 
-    pub fn open_as_if_parented<H, B>(options: WindowOpenOptions, build: B) -> RawWindowHandle
+    pub fn open_as_if_parented<H, B>(
+        options: WindowOpenOptions, build: B,
+    ) -> (RawWindowHandle, ChildWindowHandle)
```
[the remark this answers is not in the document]
BillyDM: Well, that works in the `open_as_if_parented` case, but that won't work in the `open_parented` case where we don't always get the RawWindowHandle. I suppose we could add a method to `ChildWindowHandle` that returns an `Option<HasRawWindowHandle>`.

src/window.rs
BillyDM: Hmm, I think wrl wanted to eventually have the ability to have the app accept or decline a close request (like pop-up a dialog that says "Do you want to save before you exit?" kind of thing).

Of course there are also occasions where the window *must* close, like when a DAW requests the window to close.

src/win/window.rs
BillyDM: This is not a constant because we are registering a unique ID at runtime with.
```rust
let destroy_msg_id =
                RegisterWindowMessageA("Baseview::DestroyMsg\0".as_ptr() as LPCSTR);
```
We *could* do what winit does and use the `lazy_static` crate, but I'm not sure how well that will work in a plugin context.

src/window.rs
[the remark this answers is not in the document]
BillyDM: The handle is meant to be Send. It's whole purpose is when it is dropped by the parent thread it signals the window thread to close.

src/win/window.rs
BillyDM: Although can the DAW be on a different thread?

src/window.rs
glowcoil: Yes, that ability to accept or decline would be in response to the X button being pressed, which is delivered as a particular kind of event (`WM_CLOSE` on Windows, `windowShouldClose:` on macOS, and `WM_DELETE_WINDOW` on X11). So that would make sense as a new `WindowEvent` called `ShouldClose` or `RequestClose`. That's a different situation from this method, which is the application itself telling the window to close.

src/win/window.rs
BillyDM: Hmm, I suppose so.

src/win/window.rs
glowcoil: Oh, I assumed you were just using a [`WM_USER`](https://docs.microsoft.com/en-us/windows/win32/winmsg/wm-user) event, which is what [druid-shell does](https://github.com/linebender/druid/blob/b6d389dec1aaafb239e7f0e72caac0d9de4285b6/druid-shell/src/backend/windows/window.rs#L332). It doesn't require any registration so might be simpler, but up to you.

src/win/window.rs
BillyDM: Well on the other platforms the `ParentHandle` contains two atomic bools. I figured I would just keep it consistent.

src/win/window.rs
glowcoil: On Windows and Mac, in every DAW, opening and closing the editor will happen on the same thread as the window.

src/window.rs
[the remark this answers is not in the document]
glowcoil: It will need to exist on a different thread from the window on X11 specifically, but that doesn't mean the handle itself needs to be `Send`. On all three platforms, there is a single thread on which all plugin API windowing operations happen, including requesting for the plugin to open and close its editor window.

So from the standpoint of Baseview's API, the `ChildWindowHandle` is created via `Window::open_parented()` on the windowing thread, it stays on the windowing thread for the lifetime of the editor, and then it gets dropped (or `close()` is called) on the windowing thread. The fact that specifically on X11 Baseview runs the event loop on a different thread doesn't mean that there is any need for a `ChildWindowHandle` to be sent between threads at any point.

Another reason to keep `ChildWindowHandle` `!Send` is that we want closing the window to be performed synchronously. On X11 we can do that using a synchronous channel or by joining the event loop thread, but on Windows and macOS that has to be done using a windowing API call, and that call should not be performed on a different thread from the window.

src/window.rs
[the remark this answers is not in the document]
glowcoil: Well, I think technically there's no reason we can't also provide the `RawWindowHandle` in the `open_parented` case. But I think it's fine to leave the API like this for now.

src/win/window.rs
BillyDM: Oh okay. Where does the `WM_USER` constant come from?

src/window.rs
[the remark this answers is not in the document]
BillyDM: Ah, okay.

src/win/window.rs
glowcoil: It's defined in `winapi`: https://docs.rs/winapi/0.3.9/winapi/um/winuser/constant.WM_USER.html

src/win/window.rs
```diff
@@ -51,6 +55,65 @@ unsafe fn generate_guid() -> String {
...
+impl WindowHandle {
+    pub fn close(&mut self) {
+        if let Some(hwnd) = self.hwnd.take() {
+            unsafe {
+                PostMessageW(hwnd, BV_WINDOW_MUST_CLOSE, 0, 0);
+            }
+        }
+    }
+
+    pub fn window_was_dropped(&self) -> bool {
```
glowcoil, line 75: Can this be renamed to `is_open` (with the opposite true/false values)? I think that's a lot more clear/obvious.

glowcoil (commented):
Looks good apart from these two changes!

src/window.rs
```diff
@@ -12,6 +12,8 @@ use crate::win as platform;
 #[cfg(target_os = "linux")]
 use crate::x11 as platform;
 
+pub use platform::WindowHandle;
```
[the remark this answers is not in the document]
BillyDM: Ah, okay. That makes sense.

src/x11/window.rs
```diff
@@ -13,6 +16,70 @@ use crate::{
...
+    window_dropped: Arc<AtomicBool>,
+
+    // Ensure handle is !Send
+    _phantom: PhantomData<*mut ()>,
+}
+
+impl WindowHandle {
+    pub fn close(&mut self) {
+        if let Some(_) = self.raw_window_handle.take() {
+            self.close_requested.store(true, Ordering::Relaxed);
```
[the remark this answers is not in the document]
BillyDM: Yeah. I figured I would just do the simplest solution for now so the issue with closing and reopening windows could be fixed in Linux. I'll add a TODO comment explaining the proper solution.

src/win/window.rs
BillyDM: ok

glowcoil (approved):

PR #110, opened 2022-01-17
macOS: scroll event handling

ingo-dsp:
Adds handling of mouse scrolling using the mouse wheel or touch pad for macOS. Although basic event handling for macOS had been added before in #52, it looks to me like support for this was still missing. I had a look at all other public forks of this library and could not find someone working on this.

So I looked up the API documentation for it [1] and had a look at the existing code for `mouse_move` to make a change that feels consistent with the existing code. There seem to be two kinds of scroll delta values [2] [3] [4], precise and imprecise ones. I expect the precise deltas to be used when a touchpad is used and the imprecise ones if a mouse wheel is use. But I can only test the mouse wheel scenario, as I have a desktop computer and thus no touch pad to test with.

If this is something you'd like to merge, feel free to do so. If this code needs to be improved, give me feedback and I will try to address it or just make adjustments yourself. But if you have other plans and do not want to merge this, that's fine too.

[1] https://developer.apple.com/documentation/appkit/nsresponder/1534192-scrollwheel?language=objc
[2] https://developer.apple.com/documentation/appkit/nsevent/1525758-hasprecisescrollingdeltas
[3] https://developer.apple.com/documentation/appkit/nsevent/1524505-scrollingdeltax

### DIFF
--- a/src/macos/view.rs
+++ b/src/macos/view.rs
@@ -15,7 +15,7 @@ use uuid::Uuid;
 
 use crate::MouseEvent::{ButtonPressed, ButtonReleased};
 use crate::{
-    Event, EventStatus, MouseButton, MouseEvent, Point, Size, WindowEvent, WindowInfo,
+    Event, EventStatus, MouseButton, MouseEvent, Point, ScrollDelta, Size, WindowEvent, WindowInfo,
     WindowOpenOptions,
 };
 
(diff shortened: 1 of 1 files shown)


wrl:
thanks! sorry this took so long to merge :sweat_smile:

ingo-budde:
:+1:

PR #113, opened 2022-02-04
Add error checking on X11 window creation, and fix parented X11 window creation

prokopyl:
This PR fixes creation of parented windows on X11 by using the parent's visual instead of the screen root's one, and also uses `xcb::create_window_checked` to ensure any error is caught and displayed.


robbert-vdh:
This parent visual change also seems to fix the embedding in REAPER.

PR #114, opened 2022-02-07
Create the X11 window with a depth of 32-bits

robbert-vdh:
This was supposed to fix https://github.com/glowcoil/raw-gl-context/issues/2, but it seems like setting the correct visual wasn't the magic bullet. Still, having the windows created with a 32-bit visual seems like a good idea since almost everything does that and now you can rely on the baseview window always having the same depth instead of it being dependent on the parent window.

### DIFF
--- a/src/x11/window.rs
+++ b/src/x11/window.rs
@@ -205,10 +205,39 @@ impl Window {
 
         let window_info = WindowInfo::from_logical_size(options.size, scaling);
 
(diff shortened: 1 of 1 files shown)


robbert-vdh:
I'll close this since this does not solve any problems and it triggers the same mismatching visual problem with Mesa (which just happens to work with the first visual by coincidence). I'll open a new PR after I integrated raw-gl-context back into baseview.

PR #115, opened 2022-02-07
Merge raw-gl-context into baseview to allow OpenGL contexts to be created during window creation

robbert-vdh:
As discussed on the Rust Audio Discord, this is necessary to be able to create an OpenGL context under X11. The OpenGL context configuration determines which framebuffer configuration should be used, but the framebuffer configuration in turn determines which visual should be used for the window the context will be used with. Because of that, it's not possible to create an OpenGL context for an X11 window after the fact. Because of that, the context creation from raw-gl-context's X11 module has now been split up into two parts: initializing the framebuffer config and determining a visual, and another one for actually creating the context after the window has been created. All of the other code from raw-gl-context is unchanged. The version of raw-gl-context used is from @prokopyl's PR adding more error checks on X11 (https://github.com/glowcoil/raw-gl-context/pull/15).

This is all put behind a new `opengl` feature that's disabled by default. The idea is that you can provide the OpenGL context options alongside the other window options, and if that's provided then calling `window.gl_context()` will return an `Option<&GlContext>` so you can access the OpenGL context in your window handler.

I've also taken the liberty to fix a lot of warnings. The main things remaining are a lot of unused cursor functions in the X11 implementation, and the immutable reference to mutable reference cast in the macOS implementation.

I've only tested this on Linux, so I would very much appreciate it if someone on macOS or Windows could test those implementations! On Linux it solves https://github.com/glowcoil/raw-gl-context/issues/2 for both NVIDIA drivers and Mesa, and for 0 and 8 bit alpha buffers. There's currently no example in the repo, but you can try this version of [egui-baseview](https://github.com/BillyDM/egui-baseview) that's updated to work with these changes instead:

```bash
git clone https://github.com/robbert-vdh/egui-baseview.git -b fix/update-dependencies
cargo run --example simple_demo
```

This supersedes #114.

### DIFF
--- a/src/gl/x11/errors.rs
+++ b/src/gl/x11/errors.rs
@@ -1,0 +1,131 @@
+use std::ffi::CStr;
+use std::fmt::{Debug, Formatter};
+use x11::xlib;
+
+use std::panic::AssertUnwindSafe;
+use std::sync::atomic::{AtomicPtr, Ordering};
+use std::sync::Mutex;
+
+static CURRENT_ERROR_PTR: AtomicPtr<Mutex<Option<xlib::XErrorEvent>>> =
+    AtomicPtr::new(::std::ptr::null_mut());
+
+/// A helper struct for safe X11 error handling
+pub struct XErrorHandler<'a> {
+    display: *mut xlib::Display,
+    mutex: &'a Mutex<Option<xlib::XErrorEvent>>,
+}
+
+impl<'a> XErrorHandler<'a> {
+    /// Syncs and checks if any previous X11 calls returned an error
+    pub fn check(&mut self) -> Result<(), XLibError> {
+        // Flush all possible previous errors
+        unsafe {
+            xlib::XSync(self.display, 0);
+        }
+        let error = self.mutex.lock().unwrap().take();
+
+        match error {
+            None => Ok(()),
+            Some(inner) => Err(XLibError { inner }),
+        }
+    }
+
+    /// Sets up a temporary X11 error handler for the duration of the given closure, and allows
+    /// that closure to check on the latest X11 error at any time
+    pub fn handle<T, F: FnOnce(&mut XErrorHandler) -> T>(
+        display: *mut xlib::Display,
+        handler: F,
+    ) -> T {
+        unsafe extern "C" fn error_handler(
+            _dpy: *mut xlib::Display,
+            err: *mut xlib::XErrorEvent,
+        ) -> i32 {
+            // SAFETY: the error pointer should be safe to copy
+            let err = *err;
+            // SAFETY: the ptr can only point to a valid instance or be null
+            let mutex = CURRENT_ERROR_PTR.load(Ordering::SeqCst).as_ref();
+            let mutex = if let Some(mutex) = mutex {
+                mutex
+            } else {
+                return 2;
+            };
+
+            match mutex.lock() {
+                Ok(mut current_error) => {
+                    *current_error = Some(err);
+                    0
+                }
+                Err(e) => {
+                    eprintln!(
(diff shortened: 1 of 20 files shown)


src/gl/x11/errors.rs
```diff
@@ -0,0 +1,131 @@
+use std::ffi::CStr;
+use std::fmt::{Debug, Formatter};
+use x11::xlib;
+
+use std::panic::AssertUnwindSafe;
+use std::sync::atomic::{AtomicPtr, Ordering};
+use std::sync::Mutex;
+
+static CURRENT_ERROR_PTR: AtomicPtr<Mutex<Option<xlib::XErrorEvent>>> =
```
glowcoil, line 9: I left a comment on the `raw-gl-context` PR describing how this needs to be changed: https://github.com/glowcoil/raw-gl-context/pull/15#pullrequestreview-891405163

src/win/window.rs
```diff
@@ -357,10 +371,28 @@ struct WindowState {
     handler: Box<dyn WindowHandler>,
     scale_policy: WindowScalePolicy,
     dw_style: u32,
+
+    #[cfg(feature = "opengl")]
+    gl_context: Arc<Option<GlContext>>,
```
glowcoil, line 376: This shouldn't need to be `Arc`, right? Any reason it can't just live in `WindowState`?

glowcoil (changes requested):

src/gl/x11/errors.rs
robbert-vdh: I noticed that the handler also throws away earlier errors if multiple errors occur during the `handle()` call. I think it might make more sense to at least hold on to the earliest error (since that's likely the cause of all other errors), but should the other errors be thrown away silently? Some other options would be to print to STDERR (probably not a very desirable solution) or to store multiple errors in a `Vec<XErrorEvent>` (but realistically you only care about either the first one or about none at all).

glowcoil: `handle` lets the closure check the value of the most recent error at any time during its execution (via `XErrorHandler::check`), so I don't think this is actually an issue.

src/win/window.rs
robbert-vdh: I guess it can - I just thought this was a bit cleaner. The alternative would be to replace this function:

https://github.com/RustAudio/baseview/pull/115/files/ba442a438236975cd6fffa98180e38a02730a864#diff-094de430ed1a9d02e97519cc11105495fc4e357d0ee1391f744998a8af095914R610-R613

With some nasty function that borrows from the RefCell (so you need to pray that the Win32 message handler isn't running at the same time or you get a panic) and then does some lifetime transmutations to extend the lifetime of the context beyond the RefCell borrow. So I went with an Arc instead.

glowcoil: That's fair. I think the right way to do this would be to split up the fields of `WindowState` to use multiple `Cell`s and `RefCell`s, so e.g. the `GlContext` could be borrowed immutably while the `WindowHandler` is borrowed mutably, but this is fine for now.

src/gl/x11/errors.rs
glowcoil: Thanks! One last nitpick: can you change the `Mutex` to a `Cell` or `RefCell` now that it's a thread local?

robbert-vdh: Ah yeah good one, changed it to a `RefCell`.

glowcoil (approved):

WeirdConstructor:
Thanks a lot for the work on this. It fixed my long standing issues with context creation in HexoTK (which used baseview, raw_gl_context and femtovg)!

PR #120, opened 2022-03-11
Allow taking ownership of the OpenGL context

robbert-vdh:
This is a follow up to #115. Having the window own the context seemed like a sensible approach to me since there's an `on_frame()` callback where you'd do your rendering, but some integrations like @BillyDM's [iced_baseview](https://github.com/BillyDM/iced_baseview/) do the rendering from a separate `Send + Sync + 'static` context, and you thus need to be able to move the OpenGL context into there. This change means that you'll need to move the `GlContext` into your window handle struct when first creating it by calling `window.gl_context().take()`.

### DIFF
--- a/src/macos/window.rs
+++ b/src/macos/window.rs
@@ -303,8 +303,8 @@ impl Window {
     }
 
     #[cfg(feature = "opengl")]
-    pub fn gl_context(&self) -> Option<&GlContext> {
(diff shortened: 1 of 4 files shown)


robbert-vdh:
Since this allows the context to outlive the window this isn't a great solution. We'll need to find a different way to make `iced_baseview` work.

PR #125, opened 2022-07-23
WIP: Input focus handling

gamingrobot:
When a child window is spawned by `open_parented` the KEYUP/KEYDOWN events are still sent to the parent. This makes it so the child becomes focused so key events can go to its wnd_proc function.

### DIFF
--- a/src/macos/window.rs
+++ b/src/macos/window.rs
@@ -298,6 +298,14 @@ impl Window {
         window_handle
     }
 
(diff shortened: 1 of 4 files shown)


robbert-vdh:
I'd say the correct solution would be to add callbacks to baseview to allow the application to grab and release input focus as needed. That sounds like a much better idea than grabbing input focus on mouse down and then never returning it back again.

PR #128, opened 2022-09-14
macOS: add the ability to change the mouse cursor

### LINKED ISSUE #123: Setting the cursor icon
Currently there's no way to change the mouse cursor icon from within a `WindowHandler`. Cursor icons are useful to indicate that UI elements are clickable or that you can enter text to them, but without the ability to set them plugins using baseview also tend to look very buggy in hosts like REAPER where the cursor icon may be stuck on some random non-pointer cursor when interacting with the GUI.

Baseview would not even need to allow you let you use any random cursor icon for this to be useful, just some common cursor types like a pointer, a grab handle, a caret, a crosshair, and a disabled icon would likely be sufficient.

discordance:
The feature was already there for X11, but wasn't available in MacOS and Windows.

The code is based on `winit` code mostly.
I've tested locally with a modified version of `iced_baseview` that I am working on for a project.
Not sure if adding a `set_mouse_cursor` to the window is the way to go but it works for me.

This is related to the #123 issue.


src/win/window.rs
```diff
@@ -350,7 +341,7 @@ unsafe fn register_wnd_class() -> ATOM {
         cbClsExtra: 0,
         cbWndExtra: 0,
         hIcon: null_mut(),
-        hCursor: LoadCursorW(null_mut(), IDC_ARROW),
+        hCursor: null_mut(), // If the class cursor is not NULL, the system restores the class cursor each time the mouse is moved.
```
robbert-vdh, line 344: The downside to defaulting this to `null_mut()` is that this now breaks every application that doesn't set a custom mouse cursor on startup by hiding the cursor by default. This behavior is then also different between Windows, Linux, and macOS.

I'm not quite sure what the best way to tackle this is. One option is to store the last cursor set through `Window::set_mouse_cursor()` in `WindowState`, default this value to `IDC_ARROW`, and set the cursor to that value whenever the user enters the window. Another less hacky approach is to also change the window class's cursor using `SetClassLong(hwnd, GCL_HCURSOR, cursor as i32);`. Neither is particularly pretty, but the latter approach is at least pretty foolproof. If you want, you can use this patch for this: (for some reason GitHub doesn't let you upload .patch files, so I gzipped it)

[0001-Revert-to-IDC_ARROW-as-default-window-class-cursor.patch.gz](https://github.com/RustAudio/baseview/files/10069060/0001-Revert-to-IDC_ARROW-as-default-window-class-cursor.patch.gz)

robbert-vdh (dismissed):
Rebased [here](https://github.com/robbert-vdh/baseview/tree/feature/set-mouse-cursor), tested on Windows/Wine and Linux (can also boot up a Mac if noone's tested that). Mostly LGTM functionality wise, though I have some comments on how this changes the current behavior on WIndows.

I would suggest using the same list of cursors and the same cursor assignments as winit though. And thus also dropping `MouseCursor::Pointer` again. The existing cursor enum was already based on this, so bringing everything back in line with winit's current definitions shouldn't be too difficult. The enum can be found [here](https://github.com/rust-windowing/winit/blob/bdcbd7d1f99464fe945596b43a922a9632456e44/src/window.rs#L1287-L1344), and I linked the platform-specific cursor assignments in the comments below. This way it becomes easier to switch between winit and baseview backends for GUI frameworks that support both, as the cursors will look the same on both platforms.

prokopyl (approved):

prokopyl:
I took the liberty of updating this PR and addressing the few issues that were previously raised.
Tested and works as expected!

@discordance Sorry that was quite a bit late, but thank you for this PR! :slightly_smiling_face:

PR #136, opened 2022-11-30
Support resizing windows after creation

### LINKED ISSUE #121: Resizing the window through a WindowHandle
It should be possible for a window handler to change the size of the window it's interacting with. Essentially all commercial plugins have resizable in some way (most modern plugins will either have a drag handle in the bottom right corner or they'll require you to resize the host's window, and older plugins tend to have a dropdown with preset size options). For modern plugin GUIs it would be nice to also be able to do the same thing with baseview.

robbert-vdh:
I've implemented this for Linux back in March, and I only just added Windows support. Still need to add macOS support, but I thought I'd submit a draft PR already because the changes needed for Windows support are pretty invasive. So we may want to discuss the best course of action first.

Resizing the window sends a `WM_SIZE` event and a couple other events to the window, so the window procedure and `WindowState` had to be changed to use interior mutability everywhere instead of wrapping the entire struct in a `RefCell`. This resolves most of the reentrant event call concerns from #130, but I'm not sure if it also prevents panics in that specific example  To make this work there's now a `deferred_tasks` queue for handling tasks that are the result of a call to one of the `Window` methods in an event handler that may also indirectly emit other events. Perhaps `Window::close()` should also be implemented this way.

The Linux implementation was extremely straightforward, and I'll get to the macOS one next.

This closes #121.

### DIFF
--- a/src/macos/window.rs
+++ b/src/macos/window.rs
@@ -124,6 +124,13 @@ impl Window {
     {
         let pool = unsafe { NSAutoreleasePool::new(nil) };
 
+        let scaling = match options.scale {
+            WindowScalePolicy::ScaleFactor(scale) => scale,
+            WindowScalePolicy::SystemScaleFactor => 1.0,
+        };
+
+        let window_info = WindowInfo::from_logical_size(options.size, scaling);
+
         let handle = if let RawWindowHandle::AppKit(handle) = parent.raw_window_handle() {
             handle
         } else {
@@ -144,7 +151,7 @@ impl Window {
                 .map(|gl_config| Self::create_gl_context(None, ns_view, gl_config)),
         };
 
-        let window_handle = Self::init(true, window, build);
+        let window_handle = Self::init(true, window, window_info, build);
 
         unsafe {
             let _: id = msg_send![handle.ns_view as *mut Object, addSubview: ns_view];
@@ -164,6 +171,13 @@ impl Window {
     {
         let pool = unsafe { NSAutoreleasePool::new(nil) };
 
+        let scaling = match options.scale {
+            WindowScalePolicy::ScaleFactor(scale) => scale,
+            WindowScalePolicy::SystemScaleFactor => 1.0,
+        };
+
+        let window_info = WindowInfo::from_logical_size(options.size, scaling);
+
         let ns_view = unsafe { create_view(&options) };
 
         let window = Window {
@@ -178,7 +192,7 @@ impl Window {
                 .map(|gl_config| Self::create_gl_context(None, ns_view, gl_config)),
         };
 
-        let window_handle = Self::init(true, window, build);
+        let window_handle = Self::init(true, window, window_info, build);
 
         unsafe {
             let () = msg_send![pool, drain];
@@ -254,7 +268,7 @@ impl Window {
                 .map(|gl_config| Self::create_gl_context(Some(ns_window), ns_view, gl_config)),
         };
 
-        let _ = Self::init(false, window, build);
+        let _ = Self::init(false, window, window_info, build);
 
         unsafe {
             ns_window.setContentView_(ns_view);
@@ -266,7 +280,9 @@ impl Window {
         }
     }
 
-    fn init<H, B>(parented: bool, mut window: Window, build: B) -> WindowHandle
+    fn init<H, B>(
+        parented: bool, mut window: Window, window_info: WindowInfo, build: B,
+    ) -> WindowHandle
     where
         H: WindowHandler + 'static,
         B: FnOnce(&mut crate::Window) -> H,
@@ -285,6 +301,7 @@ impl Window {
             keyboard_state: KeyboardState::new(),
             frame_timer: None,
             retain_count_after_build,
+            window_info,
             _parent_handle: parent_handle,
         }));
 
@@ -325,6 +342,8 @@ pub(super) struct WindowState {
     frame_timer: Option<CFRunLoopTimer>,
     _parent_handle: Option<ParentHandle>,
     pub retain_count_after_build: usize,
+    /// The last known window info for this window.
+    pub window_info: WindowInfo,
 }
 
 impl WindowState {
--- a/src/macos/window.rs
+++ b/src/macos/window.rs
@@ -229,10 +229,7 @@ impl Window {
 
         let rect = NSRect::new(
             NSPoint::new(0.0, 0.0),
-            NSSize::new(
-                window_info.logical_size().width as f64,
-                window_info.logical_size().height as f64,
-            ),
+            NSSize::new(window_info.logical_size().width, window_info.logical_size().height),
         );
 
         let ns_window = unsafe {
--- a/src/macos/window.rs
+++ b/src/macos/window.rs
@@ -320,7 +320,9 @@ impl Window {
     }
 
     pub fn resize(&mut self, size: Size) {
-        let size = NSSize::new(size.width, size.height);
+        // NOTE: macOS gives you a personal rave if you pass in fractional pixels here. Even though
+        //       the size is in fractional pixels.
+        let size = NSSize::new(size.width.round(), size.height.round());
 
         unsafe { NSView::setFrameSize(self.ns_view, size) };
         unsafe {
--- a/src/macos/window.rs
+++ b/src/macos/window.rs
@@ -336,7 +336,12 @@ impl Window {
         // If this is a standalone window then we'll also need to resize the window itself
         if let Some(ns_window) = self.ns_window {
             let mut frame = unsafe { NSWindow::frame(ns_window) };
+
+            // macOS wants you to manually add the size of the window decorations to the frame's
+            // size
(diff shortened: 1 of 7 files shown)


robbert-vdh:
So one thing I noticed is that the behavior for window resize notifications is different between the different platforms. On Windows the window handler would receive an event after the resize had successfully taken place, and on Linux this didn't happen. I changed the Linux implementation so it also does that. 

On Windows and macOS the implementations also sent you duplicate `Event::Window(WindowEvent::Resized(window_info))` events even if the new size is the same as the old one, so I changed the behaviors there to also only send the event if something changed.

robbert-vdh:
So the Linux and Windows versions work great. The macOS version doesn't seem to resize the view at all (standalone windows are still resized, but the contents don't resize). Apple's docs on this are extremely sparse, so if anyone with more experience programming for their platform would like to chime in I'd appreciate it a lot!

robbert-vdh:
Okay, everything _should_ work great now. I'd appreciate it if others could give this a try as well! I've implemented support for this in my [vizia fork](https://github.com/robbert-vdh/vizia) which is used by [`nih_plug_vizia`](https://github.com/robbert-vdh/nih-plug/). All plugins in that repo that use Vizia will support resizing and you can also try running `cargo run -p gain_gui_vizia --release`.

robbert-vdh:
Oh and something I have not yet done in this PR but that might improve perceived responsiveness is to draw a new frame as soon as the window resize event is received. Right now the `.on_frame()` call lags a bit behind the resize.

src/macos/window.rs
```diff
@@ -302,6 +319,35 @@ impl Window {
...
+        }
+
+        // If this is a standalone window then we'll also need to resize the window itself
+        if let Some(ns_window) = self.ns_window {
+            let mut frame = unsafe { NSWindow::frame(ns_window) };
+
+            // macOS wants you to manually add the size of the window decorations to the frame's
+            // size
+            frame.size = size;
+            let frame = unsafe { NSWindow::frameRectForContentRect_(ns_window, frame) };
```
glowcoil, line 345: This seems slightly incorrect — we're getting the frame rect, then setting its size to the content rect size, then converting it from content rect to frame rect. I think there should be a `contentRectForFrameRect:` call before setting the size.

I haven't tested this, but it seems possible for this code to end up moving the window position by the amount that the content rect is offset from the frame rect.

glowcoil (changes requested):

src/macos/window.rs
robbert-vdh: Sorry, somehow I saw your response and then promptly forgot about it until now! This does behave correctly in both embedded and standalone situations (and I saw this approach in multiple posts on Stack Overflow), but after looking at the `NSWindow` documentation for the seventh time I finally spotted a [`setContentSize`](https://developer.apple.com/documentation/appkit/nswindow/1419100-setcontentsize) which should do exactly what we want.

I tested 512c3f2393d606f91963773c73cbf3ab8de5eb57 with both NIH-plug+VIZIA and with a quick modification to the standalone that resizes when a certain key is pressed and both versions behave the same.

src/win/window.rs
```diff
@@ -61,7 +62,7 @@ const WIN_FRAME_TIMER: usize = 4242;
 
 pub struct WindowHandle {
     hwnd: Option<HWND>,
-    is_open: Arc<AtomicBool>,
+    is_open: Rc<AtomicBool>,
```
glowcoil, line 65: Can we make this flag a `Cell<bool>` instead of an `AtomicBool`?

src/win/window.rs
```diff
@@ -378,37 +419,96 @@ unsafe fn unregister_wnd_class(wnd_class: ATOM) {
...
+    handler: RefCell<Box<dyn WindowHandler>>,
     scale_policy: WindowScalePolicy,
     dw_style: u32,
 
+    /// Tasks that should be executed at the end of `wnd_proc`. This is needed to avoid mutably
+    /// borrowing the fields from `WindowState` more than once. For instance, when the window
+    /// handler requests a resize in response to a keyboard event, the window state will already be
+    /// borrowed in `wnd_proc`. So the `resize()` function below cannot also mutably borrow that
+    /// window state at the same time.
+    deferred_tasks: Rc<RefCell<VecDeque<WindowTask>>>,
```
glowcoil, line 443: An alternate design here would be to store the `WindowHandler` in an `Option`, so that `Window` can just hold a direct reference to `WindowState` itself even during the build closure when the handler doesn't exist yet. That way `deferred_tasks` wouldn't need to be in an `Rc`. But I don't think that's an obvious improvement at this point (though it will probably be cleaner once `Window` accumulates more fields), so it's up to you if you want to keep it as is.

glowcoil (changes requested):

src/win/window.rs
```diff
@@ -378,37 +419,96 @@ unsafe fn unregister_wnd_class(wnd_class: ATOM) {
...
+                let window_info = {
+                    let mut window_info = self.window_info.borrow_mut();
+                    let scaling = window_info.scale();
+                    *window_info = WindowInfo::from_logical_size(size, scaling);
+
+                    *window_info
+                };
+
+                unsafe {
+                    SetWindowPos(
```
glowcoil, line 477: I think we need to call `AdjustWindowRectEx` here, like in `open` and `WM_DPICHANGED`.

glowcoil (changes requested):

src/win/window.rs
robbert-vdh: And then wrap the entire `WindowState` in an `Rc` intead of a `Box`? The `WindowState` creates the `Window`.

src/win/window.rs
```diff
@@ -126,230 +127,270 @@ unsafe extern "system" fn wnd_proc(
...
+                    hwnd,
+                    new_rect.left as i32,
+                    new_rect.top as i32,
+                    new_rect.right - new_rect.left,
+                    new_rect.bottom - new_rect.top,
+                    SWP_NOZORDER | SWP_NOMOVE,
+                );
             }
-            WM_NCDESTROY => {
-                let window_state = Box::from_raw(window_state_ptr);
```
glowcoil, line 339: Without the `from_raw` call, the window state will result in a leak.

glowcoil (changes requested):

src/win/window.rs
robbert-vdh: That was fixed in c10ca6b40f369b0f06921024fecdd5e33d9fe514.

src/win/window.rs
robbert-vdh: Done.

src/win/window.rs
glowcoil: An `Rc` wouldn't be necessary. `Window` only exists on the stack for the duration of the build closure and event handlers, and is only passed by reference to client code, so it could safely hold an actual reference to `WindowState` (as long as care is taken when the `WindowState` is actually freed). It would involve adding a lifetime to `platform::Window`, but the public `Window` struct does already have a lifetime parameter: https://github.com/RustAudio/baseview/blob/ee156fb8840452984b5f9946635e9a35b2857c41/src/window.rs#L55

Anyway, like I said, this is just an alternate possible design, so feel free to disregard it.

robbert-vdh: Oh, I thought `Window` didn't already had a lifetime bound. That's why I thought using an `Rc` was necessary here. I've already rewritten everything using `Rc`s but I'll change that to Box+immutable references again because juggling `Rc`s in the window proc so you can clone them is kind of nasty.

robbert-vdh: I made the change. I already did something like this before I reverted it because I didn't like all the added conditional logic for the lifetime parameters on `Window`, but this does scale better than threading through the individual fields.

src/win/window.rs
robbert-vdh: Good catch! Added it in 2d1c54263913ba447bd92e2d455f7231ab7cfbd8.

src/win/window.rs
glowcoil: We could add the lifetime to all platforms' `Window` types and use `PhantomData` where necessary, but that might not be any better. Anyway, looks good to me!

src/win/window.rs
glowcoil: I think we should be passing `rect.right - rect.left` etc. to `SetWindowPos`; otherwise looks good!

robbert-vdh: Ah yeah was just mindlessly refactoring things. Done.

glowcoil (approved):

helgoboss:
@robbert-vdh Thanks, the `deferred_tasks` queue is a good way to prevent the issue mentioned in #130, I just tested it! There's one small additional thing necessary to fix the issue: An API to let baseview consumers defer tasks, because the code in #130 which has reentrancy issues is not a part of the baseview crate itself. It's part of `egui-baseview`. I will make a PR.

PR #137, opened 2022-12-10
Allow freeing NSView based on parent window closing notifications.

### LINKED ISSUE #124: macos: Support user-created wgpu surfaces.
Currently, if a user creates a wgpu surface to attach to the editor window, the window will fail to free itself since it uses `retainCount` to approximate when the window is ready to be freed:

https://github.com/RustAudio/baseview/blob/b3712638bacb3fdf2883cb5aa3f6caed0e91ac8c/src/macos/view.rs#L160-L170

I'd like to brainstorm ways to work around this. I don't fully grok the circular dependency issue yet, as I haven't repro'd it, so I want to keep that in mind.

@glowcoil do you happen to have an example of a circular dep, or hints towards how to reproduce the issue you saw?

kunalarya:
In order to use third party crates that, under the hood, may increase the ref count of the given `NSView`, we add an additional "window closing" check by querying the given `NSView` for its `NSWindow` parent, and subscribing to the `NSWindowWillCloseNotification` message.

Note that in DAWs such as Reaper, we share a parent window with other plugins, so this approach is not reliable (since you can switch between different plugins but the window persists). Those cases may require another solution.

(Mostly) fixes #124

### DIFF
--- a/src/macos/view.rs
+++ b/src/macos/view.rs
@@ -138,6 +138,10 @@ unsafe fn create_view_class() -> &'static Class {
         view_will_move_to_window as extern "C" fn(&Object, Sel, id),
     );
     class.add_method(
+        sel!(parentWindowClosing:),
+        parent_window_closing as extern "C" fn(&mut Object, Sel, id),
+    );
+    class.add_method(
         sel!(updateTrackingAreas:),
         update_tracking_areas as extern "C" fn(&Object, Sel, id),
     );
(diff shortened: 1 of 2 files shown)


glowcoil:
I am not really a fan of heuristic solutions like this. As you say, in situations like in Reaper where the `NSWindow` outlives Baseview's `NSView`, this notification won't occur; and, in the reverse situation, where the host closes the containing `NSWindow` but retains a reference to Baseview's `NSView` and then reparents it to another window, this notification will result in premature destruction of the Baseview window.

There was a discussion on the Rust Audio discord several months ago about changing Baseview's API contract so that windows may only be closed via an explicit call to `WindowHandle::close()` or `Window::close()`, which would resolve the refcount cycle issue in a more principled way. I'll write that up in more detail on #124.

PR #139, opened 2023-02-21
Macos clipboard

wtholliday:
Beginnings of copy/paste support on macOS.

### DIFF
--- a/src/clipboard.rs
+++ b/src/clipboard.rs
@@ -5,6 +5,6 @@ use crate::win as platform;
 #[cfg(target_os = "linux")]
 use crate::x11 as platform;
 
-pub fn copy_to_clipboard(data: String) {
+pub fn copy_to_clipboard(data: &str) {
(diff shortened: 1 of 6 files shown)


src/clipboard.rs
```diff
@@ -0,0 +1,11 @@
+#[cfg(target_os = "macos")]
+use crate::macos as platform;
+#[cfg(target_os = "windows")]
+use crate::win as platform;
+#[cfg(target_os = "linux")]
+use crate::x11 as platform;
+
+#[cfg(target_os = "macos")]
```
glowcoil, line 8: I think it would be preferable to add stubs with `todo!()` to the Windows and X11 backends, in the interest of keeping the public API consistent across platforms.

glowcoil (changes requested):

glowcoil (changes requested):

glowcoil (approved):

PR #140, opened 2023-02-26
Fix WM_DPICHANGED not triggering Resized event

robbert-vdh:
The window size gets changed directly in the `WM_DPICHANGED` handler, so comparing the new size against that will not work. The solution is to store the last size seen in `WM_SIZE` separately.

This fixes a regression from #136 affecting HiDPI displays+system scaling on Windows. @geom3trik was going to test this fix at some point, but everyone's forgotten about it since then and I just remembered that it still hasn't been fixed. I don't have a Windows HiDPI setup so I haven't tested this myself, but it should at least fix the `WindowEvent::Resized` not coming through when the system DPI changes. @geom3trik could you test this?

### DIFF
--- a/src/win/window.rs
+++ b/src/win/window.rs
@@ -342,11 +342,16 @@ unsafe fn wnd_proc_inner(
                     WindowInfo::from_physical_size(PhySize { width, height }, window_info.scale());
 
                 // Only send the event if anything changed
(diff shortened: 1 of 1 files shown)


geom3trik:
Tested this on my machine (Windows 11) and it works as expected.

prokopyl:
Closing this in favor of #235

PR #138, opened 2022-12-28
bump raw_window_handle to 0.5.0

saucesaft:
hi! i made some changes so that raw_window_handle can be updated to 0.5.0 (this because im using it in conjunction with [softbuffer](https://github.com/rust-windowing/softbuffer) for a personal project).

i only made the changes necessary for this pull request to work on macos since i don't have any access to another machine/os at the moment, but i'll leave this pr here in case there's initiative to update raw_window_handle in the future.

PR #134, opened 2022-11-23
Upgrade to raw-window-handle version 0.5

greatest-ape:
TODO:

- [x] Test on X11
- [x] Test on Windows
- [x] Test on macOS

https://github.com/greatest-ape/iced_baseview/tree/iced-0.5 might be useful for testing

PR #141, opened 2023-02-27
Raw window handle update

wtholliday:
Updating raw-window-handle to 0.5.

Still need to do Windows.


robbert-vdh:
This is the third PR for this (the second one being #138). What's wrong with #134?

wtholliday:
Whoops! We should merge #134

Issue #143, opened 2023-04-16
Windows: Panic "already borrowed" when reentering wnd_proc due to mouse move or timer


helgoboss:
Reentering wnd_proc can happen, e.g. when opening the system web browser in an event handler via webbrowser crate.

Originally I made PR #130 for this issue, but since PR #136 has been merged, there's a more elegant way to fix the issue. I will file a new PR.

PR #144, opened 2023-04-16
Windows: Add public defer method

helgoboss:
This solves #143 in a better way than #130 did. It lets downstream crates defer tasks until the end of the current main loop cycle in order to prevent "already borrowed" errors because of multiple mutable borrows of the window state.

I added this method for Windows only because this is the only OS where baseview uses RefCells and the only OS for which the `deferred_task` queue has been implemented (in #136).


prokopyl:
Full re-entrancy support has been implemented in #267 , so I believe this isn't needed anymore.

PR #146, opened 2023-04-16
Add support for Windows 7 and 8

### LINKED ISSUE #145: Add support for Windows 7 and 8
A few users are still on Windows 7 and 8. Currently, plug-ins which use baseview don't even scan on these platforms because `GetDpiForWindow` and `SetProcessDpiAwarenessContext` are not available there. I will make a PR which uses lazy loading of those functions.

helgoboss:
This implements #145.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ nix = "0.22.0"
 [target.'cfg(target_os="windows")'.dependencies]
 winapi = { version = "0.3.8", features = ["libloaderapi", "winuser", "windef", "minwindef", "guiddef", "combaseapi", "wingdi", "errhandlingapi"] }
 uuid = { version = "0.8", features = ["v4"], optional = true }
+libloading = "0.8.0"
 
 [target.'cfg(target_os="macos")'.dependencies]
 cocoa = "0.24.0"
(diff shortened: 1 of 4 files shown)


robbert-vdh:
Does this require adding a dependency on `libloading`? I would just use `LoadLibrary` and `GetProcAddress` since we already depend on the winapi crate anyways (which should probably be replaced with the `windows` crate at some point), and then store the function pointers in a `OnceCell` (which is now in the standard library) so they're always there without having to look them up again each time you call them.

I proposed doing this in the Discord a couple months back but no one seemed to be particularly enthusiastic about having to maintain WIndows 7 support since it may break over time if you don't periodically test on it, and Windows 7 and 8 are end of life.

helgoboss:
It doesn't strictly need `libloading`, I just thought it's easier to rely on the abstraction. But yes, I guess I could change it to use `LoadLibrary` and `GetProcAddress` directly.

I'm also not really inclined to test this regularly. But eagerly linking to the library functions completely rules out the possiblity to run this on older Windows versions ... I find this a bit restrictive for a general-purpose lib, especially when it's only because of 2 functions and the solution is rather simple. Maybe it's okay if it's not regularly tested as long as the the README doesn't actively advertise that it supports old Windows versions?

Some audio folks are quite conservative when it comes to the OS. There was a user wondering why ReaLearn doesn't work on his syststem (https://github.com/helgoboss/realearn/issues/791). That's why I added it in the first place.

prokopyl:
I'll be closing this in favor of #268, which addresses the issues raised by @robbert-vdh

PR #147, opened 2023-05-21
add support for CursorEntered and CursorLeft events in Windows OS

Amarantheum:
A simple change that uses [TrackMouseEvent](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-trackmouseevent)  to determine if the cursor has left the window. Uses WM_MOUSEMOVE to determine if the mouse has entered back into the window. Adds additional state `mouse_in_window` to the `WindowState` struct to determine if the cursor is over the window. This is initialized to true since TrackMouseEvent will immediately register a WM_MOUSELEAVE message if the cursor is not in the window.

Mostly adding this since it's useful for me so it may be useful for others.


prokopyl:
Closing this, since it turns out it has already been implemented/merged. :slightly_smiling_face:

PR #148, opened 2023-06-05
Drag and drop

ilmai:
Only implemented for Windows for now

### DIFF
--- a/src/win/drop_target.rs
+++ b/src/win/drop_target.rs
@@ -1,0 +1,266 @@
+use std::ffi::OsString;
+use std::mem::transmute;
+use std::os::windows::prelude::OsStringExt;
+use std::ptr::null_mut;
+use std::rc::{Weak, Rc};
+
+use winapi::Interface;
+use winapi::shared::guiddef::{REFIID, IsEqualIID};
+use winapi::shared::minwindef::{DWORD, WPARAM};
+use winapi::shared::ntdef::{HRESULT, ULONG};
+use winapi::shared::windef::POINTL;
+use winapi::shared::winerror::{S_OK, E_NOINTERFACE, E_UNEXPECTED};
+use winapi::shared::wtypes::DVASPECT_CONTENT;
+use winapi::um::objidl::{IDataObject, FORMATETC, TYMED_HGLOBAL, STGMEDIUM};
+use winapi::um::oleidl::{IDropTarget, IDropTargetVtbl, DROPEFFECT_COPY, DROPEFFECT_MOVE, DROPEFFECT_LINK, DROPEFFECT_SCROLL, DROPEFFECT_NONE};
+use winapi::um::shellapi::DragQueryFileW;
+use winapi::um::unknwnbase::{IUnknownVtbl, IUnknown};
+use winapi::um::winuser::CF_HDROP;
+
+use crate::{Point, DropData, MouseEvent, Event, EventStatus, DropEffect, PhyPoint};
+
+use super::WindowState;
+
+// These function pointers have to be stored in a (const) variable before they can be transmuted
+const DRAG_ENTER_PTR: unsafe extern "system" fn(this: *mut IDropTarget, pDataObj: *const IDataObject, grfKeyState: DWORD, pt: POINTL, pdwEffect: *mut DWORD) -> HRESULT = DropTarget::drag_enter;
+const DRAG_OVER_PTR: unsafe extern "system" fn(this: *mut IDropTarget, grfKeyState: DWORD, pt: POINTL, pdwEffect: *mut DWORD) -> HRESULT = DropTarget::drag_over;
+const DROP_PTR: unsafe extern "system" fn(this: *mut IDropTarget, pDataObj: *const IDataObject, grfKeyState: DWORD, pt: POINTL, pdwEffect: *mut DWORD) -> HRESULT = DropTarget::drop;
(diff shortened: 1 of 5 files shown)


src/win/window.rs
```diff
@@ -767,3 +789,244 @@ unsafe impl HasRawWindowHandle for Window<'_> {
...
+
+    // These are cached since DragOver and DragLeave callbacks don't provide them,
+    // and handling drag move events gets awkward on the client end otherwise
+    drag_position: Point,
+    drop_data: DropData,
+}
+
+impl DropTarget {
+    fn new(window_state: *mut WindowState) -> Self {
+        let vtbl = Arc::new(IDropTargetVtbl {
```
glowcoil, line 808: You can simplify this by getting rid of the `Arc` and just using a const or static for the vtable.

src/win/window.rs
```diff
@@ -767,3 +789,244 @@ unsafe impl HasRawWindowHandle for Window<'_> {
...
+
+    unsafe extern "system" fn query_interface(
+        this: *mut IUnknown,
+        riid: REFIID,
+        ppvObject: *mut *mut winapi::ctypes::c_void,
+    ) -> HRESULT
+    {
+        if IsEqualIID(&*riid, &IUnknown::uuidof()) || IsEqualIID(&*riid, &IDropTarget::uuidof()){
+            Self::add_ref(this);
+            *ppvObject = unsafe { transmute(this) };
```
glowcoil, line 921: `transmute` shouldn't be necessary here; this can just be a regular pointer cast with `as`.

src/win/window.rs
```diff
@@ -767,3 +789,244 @@ unsafe impl HasRawWindowHandle for Window<'_> {
 pub fn copy_to_clipboard(data: &str) {
     todo!()
 }
+
```
glowcoil, line 792: I think it would be nice to have all of the `IDropTarget` implementation in a separate file.

src/win/window.rs
```diff
@@ -767,3 +789,244 @@ unsafe impl HasRawWindowHandle for Window<'_> {
...
+#[repr(C)]
+pub struct DropTarget {
+    base: IDropTarget,
+    vtbl: Arc<IDropTargetVtbl>,
+
+    window_state: *mut WindowState,
```
glowcoil, line 798: This worries me. I think if `DropTarget` has a reference back to `WindowState`, it should use an `Rc` or `Weak` rather than just a raw pointer.

src/win/window.rs
```diff
@@ -711,7 +725,15 @@ impl Window<'_> {
                 None
             };
 
-            SetWindowLongPtrW(hwnd, GWLP_USERDATA, Box::into_raw(window_state) as *const _ as _);
+            let window_state_ptr = Box::into_raw(window_state);
+            let drop_target = Arc::new(DropTarget::new(window_state_ptr));
```
glowcoil, line 729: It should be fine to use `Rc` rather than `Arc`, as everything here is single-threaded.

glowcoil (changes requested):

src/win/window.rs
```diff
@@ -767,3 +789,244 @@ unsafe impl HasRawWindowHandle for Window<'_> {
...
+                    EventStatus::AcceptDrop(DropEffect::Move) => *pdwEffect = DROPEFFECT_MOVE,
+                    EventStatus::AcceptDrop(DropEffect::Link) => *pdwEffect = DROPEFFECT_LINK,
+                    EventStatus::AcceptDrop(DropEffect::Scroll) => *pdwEffect = DROPEFFECT_SCROLL,
+                    _ => *pdwEffect = DROPEFFECT_NONE,
+                }        
+            } 
+        }
+    }
+
+    fn parse_coordinates(&mut self, pt: *const POINTL) {
```
glowcoil, line 852: I think a safer solution to this (short of fixing `winapi` or switching to `windows-sys`) would be to write `drag_enter`, `drag_over`, and `drop` with the correct type signatures and then use `transmute` when placing the function pointers in the vtable.

src/win/drop_target.rs
```diff
@@ -0,0 +1,266 @@
...
+const DRAG_ENTER_PTR: unsafe extern "system" fn(this: *mut IDropTarget, pDataObj: *const IDataObject, grfKeyState: DWORD, pt: POINTL, pdwEffect: *mut DWORD) -> HRESULT = DropTarget::drag_enter;
+const DRAG_OVER_PTR: unsafe extern "system" fn(this: *mut IDropTarget, grfKeyState: DWORD, pt: POINTL, pdwEffect: *mut DWORD) -> HRESULT = DropTarget::drag_over;
+const DROP_PTR: unsafe extern "system" fn(this: *mut IDropTarget, pDataObj: *const IDataObject, grfKeyState: DWORD, pt: POINTL, pdwEffect: *mut DWORD) -> HRESULT = DropTarget::drop;
+const DROP_TARGET_VTBL: IDropTargetVtbl = IDropTargetVtbl {
+    parent: IUnknownVtbl {
+        QueryInterface: DropTarget::query_interface,
+        AddRef: DropTarget::add_ref,
+        Release: DropTarget::release,
+    },
+    DragEnter: unsafe { transmute(DRAG_ENTER_PTR) },
```
glowcoil, line 34: Probably worth adding a comment that explicitly states why these transmutes are necessary.

glowcoil (approved):
LGTM!

PR #150, opened 2023-10-02
Add logic for CursorEntered/CursorLeft on Windows

Fredemus:
Necessary to fix https://github.com/vizia/vizia/issues/407, which is a bug that locks up the GUI when you press down on a mouse button inside the window and release the button outside the window.

### DIFF
--- a/src/win/window.rs
+++ b/src/win/window.rs
@@ -2,23 +2,23 @@ use winapi::shared::guiddef::GUID;
 use winapi::shared::minwindef::{ATOM, FALSE, LPARAM, LRESULT, UINT, WPARAM};
 use winapi::shared::windef::{HWND, RECT};
 use winapi::um::combaseapi::CoCreateGuid;
-use winapi::um::ole2::{RegisterDragDrop, OleInitialize, RevokeDragDrop};
+use winapi::um::ole2::{OleInitialize, RegisterDragDrop, RevokeDragDrop};
 use winapi::um::oleidl::LPDROPTARGET;
 use winapi::um::winuser::{
     AdjustWindowRectEx, CreateWindowExW, DefWindowProcW, DestroyWindow, DispatchMessageW,
     GetDpiForWindow, GetMessageW, GetWindowLongPtrW, LoadCursorW, PostMessageW, RegisterClassW,
     ReleaseCapture, SetCapture, SetProcessDpiAwarenessContext, SetTimer, SetWindowLongPtrW,
-    SetWindowPos, TranslateMessage, UnregisterClassW, CS_OWNDC, GET_XBUTTON_WPARAM, GWLP_USERDATA,
-    IDC_ARROW, MSG, SWP_NOMOVE, SWP_NOZORDER, WHEEL_DELTA, WM_CHAR, WM_CLOSE, WM_CREATE,
-    WM_DPICHANGED, WM_INPUTLANGCHANGE, WM_KEYDOWN, WM_KEYUP, WM_LBUTTONDOWN, WM_LBUTTONUP,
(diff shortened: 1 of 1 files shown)


robbert-vdh (changes requested):
I think this can be implemented without any additional state tracking or heuristics.

Take a look at the documentation of [`TrackMouseEvent`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-trackmouseevent). You can use it to get [`WM_MOUSEHOVER`](https://learn.microsoft.com/en-us/windows/desktop/inputdev/wm-mousehover) events when the cursor enters the client area, and [`WM_MOUSELEAVE`](https://learn.microsoft.com/en-us/windows/desktop/inputdev/wm-mouseleave) events when the cursor leaves the client area. You need to set both the `TME_HOVER` and the `TME_LEAVE` flags in the `TRACKMOUSEEVENT` struct for that to work, and you need to repeat the call any time you receiver either of the two window messages. So the idea is that you call it after creating the window, and you then repeat the exact same call when handling those two window messages.

Fredemus:
I first tried to do it like that, but `WM_MOUSEHOVER` ends up firing every time the mouse moves inside the window. This way it only happens the first time the mouse enters the window, which lines up with how it works on x11.

robbert-vdh:
In that case you'd be able to just only set `TME_HOVER` when initially creating the window and when handling `WM_MOUSELEAVE`, and only set `TME_LEAVE` when handling `WM_MOUSEENTER`, no? If that works that sounds like a better solution than approximating it by tracking mouse moves.

Fredemus:
> `WM_MOUSEENTER`

For some reason there is no `WM_MOUSEENTER`, only `WM_MOUSEHOVER`

robbert-vdh:
Sorry yeah I meant `WM_MOUSEHOVER`.

Fredemus:
Ah okay.  For reasons I don't really understand just sending `TrackMouseEvent` with `TME_HOVER` in `WM_MOUSELEAVE` means the event never fires, no matter if I set either `TME_HOVER` or `TME_LEAVE` initially, or both. I also tried to change `dwHoverTime` in case the system default was weird. It seemingly has to happen in `WM_MOUSEMOVE`. 

I digged up the [documentation I based this on](https://learn.microsoft.com/en-us/windows/win32/learnwin32/other-mouse-operations#mouse-tracking-events-hover-and-leave).

src/win/window.rs
```diff
@@ -175,22 +175,51 @@ unsafe fn wnd_proc_inner(
...
             let mut window = window_state.create_window();
             let mut window = crate::Window::new(&mut window);
+            let mut handler = window_state.handler.borrow_mut();
+            let handler = handler.as_mut().unwrap();
+
+            let mut mouse_was_outside_window = window_state.mouse_was_outside_window.borrow_mut();
+            if *mouse_was_outside_window {
+                // this makes Windows track whether the mouse leaves the window.
+                // When the mouse leaves it results in a `WM_MOUSELEAVE` event.
+                let track_mouse = Rc::new(TRACKMOUSEEVENT {
```
robbert-vdh, line 185: If there's no other way then I'd at least suggest removing this allocation. That is not needed.

robbert-vdh (changes requested):

src/win/window.rs
Fredemus: Sorry, somehow missed the notification. Fixed

robbert-vdh (approved):
I don't know how to feel about the Win32 API for this, but I guess this is the way to do it. :sweat_smile: 

LGTM.

PR #156, opened 2023-12-12
Remove RawWindowHandleWrapper

glowcoil:
RawWindowHandleWrapper is only necessary since GlContext::create takes an `impl HasRawWindowHandle` argument, but GlContext::create is an internal unsafe API and it can just take a RawWindowHandle directly.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -21,12 +21,6 @@ pub struct WindowHandle {
     phantom: PhantomData<*mut ()>,
 }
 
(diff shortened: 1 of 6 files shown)


src/window.rs
```diff
@@ -21,12 +21,6 @@ pub struct WindowHandle {
     phantom: PhantomData<*mut ()>,
 }
 
-/// Quick wrapper to satisfy [HasRawWindowHandle], because of course a raw window handle wouldn't
```
Engid, line 24: sad to see this comment go 😆

Engid (approved):
Great cleanup, thanks!

PR #157, opened 2023-12-16
macOS: use interior mutability for WindowState

glowcoil:
Forming mutable references to the WindowState is unsound given the possibility of reentrant calls to NSView methods. Instead, form only immutable references to the WindowState and wrap mutable fields in Cell and RefCell.

Follow-up work should use try_borrow_mut instead of borrow_mut to avoid panicking in the case of reentrant calls.


Engid (approved):
Thanks for the soundness fix. Looks great 👍

PR #164, opened 2023-12-17
Fix window cleanup logic on macOS

glowcoil:
Instead of trying to detect when to clean up the window based on the `NSView`'s retain count, require window cleanup to be initiated explicitly via `Window::close`, `WindowHandle::close`, or `[NSWindowDelegate windowShouldClose:]` (in non-parented mode; called when the user clicks the "X" button). This fixes the leaks and use-after-frees that can be caused by the inherent unreliability of the retain count logic.

As discussed in #153, this change essentially means that the `NSView` created by Baseview will not be suitable as the top-level view for an Audio Unit, since the Baseview API now requires that child windows be cleaned up by an explicit call to `WindowHandle::close`, and the only reliable signal for cleaning up an Audio Unit view is a call to `[NSView dealloc]`. However, this does not mean that Baseview cannot be used in the context of an Audio Unit; it just means that plugin frameworks must implement a compatibility layer with a wrapper `NSView` (which is the approach taken by JUCE).

In order to implement this change:

- `WindowState` is stored in an `Rc` rather than a `Box`.
- `WindowHandle` holds an `Rc<WindowState>` so that `WindowHandle::close` can directly invoke window cleanup logic.
- Since the window can be closed during an event handler, `WindowState::from_view` now returns a clone of the `Rc<WindowState>` held by the `NSView` to ensure that it lives until the end of an event handler.
- In the non-parented case, the `NSView` is set as the window delegate, which allows it to receive the `windowShouldClose:` call when the user clicks the "X" button, upon which it will dispatch the `WillClose` event and initiate window cleanup logic.
- `Window::open_parented` and `open_blocking` no longer `release` the `NSView` immediately after attaching it. Instead, the `NSView` is released as part of the cleanup logic in `WindowInner::close`.
- `Window::resize` now checks if the window is open to avoid using the `NSView` after releasing it.
- The overridden `release` method, the `retain_count_after_build` field, the `ParentHandle` struct, and the `close_requested` flag have all been removed.

### DIFF
--- a/src/macos/view.rs
+++ b/src/macos/view.rs
@@ -134,7 +134,10 @@ unsafe fn create_view_class() -> &'static Class {
         accepts_first_mouse as extern "C" fn(&Object, Sel, id) -> BOOL,
     );
 
-    class.add_method(sel!(release), release as extern "C" fn(&mut Object, Sel));
+    class.add_method(
+        sel!(windowShouldClose:),
+        window_should_close as extern "C" fn(&Object, Sel, id) -> BOOL,
+    );
     class.add_method(sel!(dealloc), dealloc as extern "C" fn(&mut Object, Sel));
     class.add_method(
         sel!(viewWillMoveToWindow:),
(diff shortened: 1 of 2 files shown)


Engid:
@glowcoil Beautiful code and great detailed descriptions here! thank you. I just have a basic question so please forgive me: Were you able to change the `Arc`'s in `window.rs` to an `Rc` because as you described
> In the non-parented case, the NSView is set as the window delegate, which allows it to receive the windowShouldClose
..? In other words, how can I tell that its safe to switch to `Rc`? Thank you.

Engid (approved):
Optimistically approving but hoping to hear back from @maxjvh with some QA testing confirmation

glowcoil:
@Engid The `Arc`s and atomics were actually never necessary in the first place, since the entire macOS backend has always been completely single-threaded. I think they were just copied unchanged from the X11 backend, which is multithreaded.

glowcoil:
Tested this PR using the `gain_gui_vizia` example plugin in the [NIH-plug repository](https://github.com/robbert-vdh/nih-plug/). On both the current `master` branch (https://github.com/robbert-vdh/nih-plug/commit/e211029979a4a4aaece4e0bc3362c312f4ef92e4) and the `vizia-update` branch (https://github.com/robbert-vdh/nih-plug/commit/cd5b2592a80c2981f038c36fd0c530771a2d7414), when running in Bitwig on macOS, the `WindowHandler` does not get dropped when the editor window is closed. After updating Vizia to use this branch of Baseview, the `WindowHandler` is successfully dropped when the editor window is closed.

greatest-ape:
Excellent!

PR #165, opened 2023-12-18
Remove obsolete roadmap from README

micahrj:


maxjvh:
perhaps "roadmap" is not a very good name for this, but i think a list of features and their support on all the different platforms would be nice to have. but this list is also missing some features like drag&drop (which is only supported on Windows and macOS afaik).

Engid:
Yeah might be good to change it to a feature list

robbert-vdh:
I agree. I think getting rid of it would also be fine, but there have been some recent PRs that introduced features for some platforms but not for others.

glowcoil:
I'll replace it with a feature list. I agree that it would be good to track what still needs to be implemented on different platforms; I just think it would be a better fit for GitHub issues (or another task tracking system) than the project README.

prokopyl:
I completely missed that PR somehow, but the roadmap has already been removed from the README so I'll close this. :slightly_smiling_face:

Issue #169, opened 2024-03-18
key events not working in Reaper on macOS


maxjvh:
[the NSView's keyDown/keyUp methods](https://github.com/RustAudio/baseview/blob/master/src/macos/view.rs#L190-L191) are never getting called in Reaper when there's keyboard input with the plugin window in focus. neither turning on "Send all keyboard input to plugin" or opening the UI in its own window changed the situation.

curiously, the events are still consumed on some level as Reaper is also not receiving keyboard events when the plugin window is in focus (space doesn't toggle playback, for example).

i'm running macOS 13.5.2 and Reaper 6.43 and using baseview with nih_plug

maxjvh:
wondering if the problem actually lies within nih_plug: the VST3 [`onKeyDown` implementation](https://github.com/robbert-vdh/nih-plug/blob/master/src/wrapper/vst3/view.rs#L343-L350) returns `kResultOk`, but according to the [API docs](https://steinbergmedia.github.io/vst3_doc/base/classSteinberg_1_1IPlugView.html#a759b576f046e699c84dc07d579600b1b) it should return either `kResultTrue` or `kResultFalse` depending on whether the key is consumed. perhaps `kResultOk` is interpreted similarly to `kResultTrue`?

i have a feeling that Reaper might be using these VST3 key handlers more pedantically than most hosts

maxjvh:
returning `kResultFalse` from the VST3 key handlers indeed allows the key events to flow to the plugin! but now the problem is that no key events are being consumed from Reaper's perspective, so e.g. the spacebar and enter keys will always get passed to Reaper. to fix this, baseview would need to have some way to signal to the VST3 key handlers about when to consume/ignore key events, but not sure what that would require.

peastman:
I'm blocked by this same problem.  I'm using NIH-Plug and implementing the UI with egui.  Key input doesn't work in either the VST3 or CLAP versions of the plugin.  Is there any possibility of this getting fixed?

peastman:
This problem can be fixed by the code change suggested in #202.

Issue #152, opened 2023-10-21
Add a `set_focus(bool)` interface for getting and losing keyboard focus


Engid:
This is a compilation of discussions from the discord chat from @ilmai et al. 

(This is a placeholder for me to edit with more details once I get another minute)

Engid:
Per ilmai: 
> For reference here's what I'm doing on Mac with my own windowing abstraction experiment that I'm using for slint. It's not really production-ready but seems to work as expected based on some testing. https://github.com/ilmai/plugin-things/blob/main/plugin-canvas/src/platform/mac/view.rs#L255

> The boolean is checked here https://github.com/ilmai/plugin-things/blob/main/plugin-canvas/src/platform/mac/view.rs#L66
Windows and Linux have similar logic

> And `set_input_focus()` is called from here https://github.com/ilmai/plugin-things/blob/main/nih_plug_slint/src/window_adapter.rs#L318

dathinaios:
(Dionysis from the Rust Audio discord here 🙂👋)

So, I am giving this a go again. I am reading the `baseview` code and follow it ok but we need more clarity on what needs to be implemented. Focusing on Mac for now but if we get the interface down it should not be too difficult to implement for other platforms (famous last words 😁).

I had a look at @ilmai 's code from the link above and I can not see where the `input_focus` boolean interacts with `NSWindow`. From a discord conversation @ilmai has said:

> you don't really need to set input focus yourself like that, but the GUI crate (vizia in this case) asking for input focus should dictate if keyboard events are passed to the parent window or not

That sounds to me as if `baseview` is not handling the focus but simply storing the boolean and then it is a matter of calling the focus method from Vizia (or other GUI) which I assume already has access to the `NSWindow`.

Is this correct? 🤔

PR #170, opened 2024-03-19
Implement input focus for Mac & Windows

dathinaios:
As per https://github.com/RustAudio/baseview/issues/152. Still not working for VST but that appears to be due to a separate issue (see https://github.com/RustAudio/baseview/issues/169). I tested it in a `nih_plug` CLAP plugin in Reaper. I had to modify `vizia_baseview` to call `set_input_focus` when the alt-click event is triggered. Will also commit those changes there if this is accepted as a solution.

### DIFF
--- a/src/macos/window.rs
+++ b/src/macos/window.rs
@@ -281,7 +281,7 @@ impl<'a> Window<'a> {
         self.inner.close();
     }
 
-    pub fn has_input_focus(&mut self) -> bool {
+    pub fn has_focus(&mut self) -> bool {
         unsafe {
             let view = self.inner.ns_view.as_mut().unwrap();
             let window: id = msg_send![view, window];
@@ -293,7 +293,7 @@ impl<'a> Window<'a> {
         }
     }
 
-    pub fn set_input_focus(&mut self) {
+    pub fn focus(&mut self) {
         unsafe {
             let view = self.inner.ns_view.as_mut().unwrap();
             let window: id = msg_send![view, window];
--- a/src/macos/window.rs
+++ b/src/macos/window.rs
@@ -7,7 +7,7 @@ use cocoa::appkit::{
     NSApp, NSApplication, NSApplicationActivationPolicyRegular, NSBackingStoreBuffered,
     NSPasteboard, NSView, NSWindow, NSWindowStyleMask,
 };
-use cocoa::base::{id, nil, NO, YES};
+use cocoa::base::{id, nil, BOOL, NO, YES};
 use cocoa::foundation::{NSAutoreleasePool, NSPoint, NSRect, NSSize, NSString};
 use core_foundation::runloop::{
     CFRunLoop, CFRunLoopTimer, CFRunLoopTimerContext, __CFRunLoopTimer, kCFRunLoopDefaultMode,
@@ -44,6 +44,7 @@ impl WindowHandle {
     pub fn is_open(&self) -> bool {
         self.state.window_inner.open.get()
     }
+
 }
 
 unsafe impl HasRawWindowHandle for WindowHandle {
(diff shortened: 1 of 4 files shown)


glowcoil:
The Windows and Linux builds will fail without definitions of `has_input_focus` and `set_input_focus` on their `platform::Window` types. Please add some stub definitions with `unimplemented!()` to the Windows and Linux backends.

src/macos/window.rs
```diff
@@ -280,6 +281,27 @@ impl<'a> Window<'a> {
...
     }
 
+    pub fn has_input_focus(&mut self) -> bool {
+        unsafe {
+            let view = self.inner.ns_view.as_mut().unwrap();
+            let window: id = msg_send![view, window];
+            if window == nil { return false; };
+            let first_responder: id = msg_send![window, firstResponder];
+            let is_focused: BOOL = msg_send![view, isEqual: first_responder];
+            is_focused == YES
```
maxjvh, line 291: might also want to check if the window is the key window:
```rs
let is_key_window: BOOL = msg_send![window, isKeyWindow];
is_key_window == YES && is_focused == YES
```
an `NSView` can be the first responder of an `NSWindow` even when the window doesn't have focus

dathinaios: That's easy to add thanks. But are we sure it is necessary in the context of `baseview`? I am still unclear about the windows and views especially when we call `open_parented` in the context of a DAW.

maxjvh: i'm not aware of the specific case this method was made for, but if it's intended to be a generic way to check whether a view has keyboard focus, then checking the window's key status is probably necessary. for example, some library using baseview might poll this field and get a false positive when the `NSView` has first responder status but the `NSWindow` is no longer the key window.

dathinaios: Ok, I will have a look. I wonder then if I should revert what I removed [here](https://github.com/RustAudio/baseview/commit/be6c7e9cd56d82813122a7ecae0d51a5907ec59d), where I was also making the window the key window on 'set_input_focus'.

maxjvh: guess so, but i still don't quite understand the use-case for explicitly setting input focus. correct me if i'm wrong, but when clicking on an `NSView`, doesn't it automatically become `firstResponder` and its window the key window? when and why do you need to explicitly do this in vizia?

dathinaios: If you search on the discord you will find all the occurrences where people had issues with this. For many DAWs there was no way to manually enter a value for a parameter. The view will appear focused but pressing keys on the keyboard would have no effect. It seems to be a common issue related to DAWs and plugins presumably because the focus is kept by the arrangement window of the DAW so that shortcuts will work (space for instance to play).

src/win/window.rs
```diff
@@ -770,6 +770,17 @@ impl Window<'_> {
         }
     }
 
+    pub fn has_input_focus(&mut self) -> bool {
+        let foreground_window = unsafe { GetForegroundWindow() };
```
glowcoil, line 774: I believe this should be `GetFocus`, not `GetForegroundWindow`.

`Get`/`SetForegroundWindow`, `Get`/`SetActiveWindow`, and `Get`/`SetFocus` all do slightly different things. My understanding from the documentation is as follows: Within a particular thread, there is a single top-level window which is considered "active" (which basically corresponds to which window from that thread is drawn in the front and has its title bar highlighted), as well as a single top-level or child window which is considered "focused" (which is the window that receives keyboard input). These can be changed independently, so a given top-level window can be both the "active" and "focused" window, or it can be the "active" window while one of its child windows is the "focused" window. Then, across all processes in the entire system, there is a single thread which is considered the "foreground" thread (which basically corresponds to which *application* is currently focused). So, `GetForegroundWindow` gets the active window for the foreground thread (which can be from a different process), and `SetForegroundWindow` sets both the foreground thread and the active window for that thread; `GetActiveWindow` and `SetActiveWindow` get and set the active window for the *current* thread, ignoring whether it is the foreground thread or not; and `GetFocus` and `SetFocus` get and set the *focused* window for the current thread, also ignoring whether it is the foreground thread or not.

Given all that, I think that `GetFocus` and `SetFocus` are the functions we want, since our plugin window is not a top-level window and what we care about is the keyboard focus.

Sources:
- Raymond Chen blog post, ["Eventually, nothing is special any more"](https://devblogs.microsoft.com/oldnewthing/20081006-00/?p=20643)
- The ["Window Features"](https://learn.microsoft.com/en-us/windows/win32/winmsg/window-features) page in the Win32 docs, specifically the ["Foreground and Background Windows"](https://learn.microsoft.com/en-us/windows/win32/winmsg/window-features#foreground-and-background-windows) and ["Active Window"](https://learn.microsoft.com/en-us/windows/win32/winmsg/window-features#active-window) sections

src/window.rs
```diff
@@ -102,6 +102,14 @@ impl<'a> Window<'a> {
         self.window.set_mouse_cursor(cursor);
     }
 
+    pub fn has_input_focus(&mut self) -> bool {
```
glowcoil, line 105: This is a nitpick, but I think these methods could just be named `has_focus` and `focus`.

src/win/window.rs
```diff
@@ -770,6 +770,17 @@ impl Window<'_> {
         }
     }
 
+    pub fn has_focus(&mut self) -> bool {
+        let foreground_window = unsafe { GetFocus() };
```
glowcoil, line 774: Do you mind renaming this variable to `focused_window`? `foreground_window` is a little misleading.

dathinaios: Ah of course. Missed that.

glowcoil (approved):

PR #171, opened 2024-03-19
add support for Window::Focused/Unfocused events on macOS

httnn:
trigger `WindowEvent::Focused` and `WindowEvent::Unfocused` events when the plugin window gains/loses focus. implemented by adding observers to `NSNotificationCenter::defaultCenter()` that listen to `NSWindowDidBecomeKeyNotification` and `NSWindowDidResignKeyNotification` notifications on the `NSViews`' window.

tested and confirmed to work in Live, Bitwig, FL Studio, Reaper and AudioPluginHost.

### DIFF
--- a/src/macos/view.rs
+++ b/src/macos/view.rs
@@ -2,7 +2,7 @@ use std::ffi::c_void;
 
 use cocoa::appkit::{NSEvent, NSFilenamesPboardType, NSView, NSWindow};
 use cocoa::base::{id, nil, BOOL, NO, YES};
-use cocoa::foundation::{NSArray, NSPoint, NSRect, NSSize, NSUInteger};
+use cocoa::foundation::{NSArray, NSPoint, NSRect, NSSize, NSString, NSUInteger};
 
 use objc::{
     class,
@@ -94,6 +94,18 @@ macro_rules! add_simple_keyboard_class_method {
     };
 }
 
+unsafe fn register_notification(observer: id, notification_name: &'static str, object: id) {
+    let notification_center: id = msg_send![class!(NSNotificationCenter), defaultCenter];
+
+    let _: () = msg_send![
+        notification_center,
+        addObserver:observer
+        selector:sel!(handleNotification:)
+        name:NSString::alloc(nil).init_str(notification_name)
+        object:object
+    ];
+}
+
 pub(super) unsafe fn create_view(window_options: &WindowOpenOptions) -> id {
     let class = create_view_class();
 
@@ -102,6 +114,9 @@ pub(super) unsafe fn create_view(window_options: &WindowOpenOptions) -> id {
     let size = window_options.size;
 
     view.initWithFrame_(NSRect::new(NSPoint::new(0., 0.), NSSize::new(size.width, size.height)));
+
+    register_notification(view, "NSWindowDidBecomeKeyNotification", nil);
+    register_notification(view, "NSWindowDidResignKeyNotification", nil);
 
     let _: id = msg_send![
         view,
@@ -185,6 +200,10 @@ unsafe fn create_view_class() -> &'static Class {
         dragging_updated as extern "C" fn(&Object, Sel, id) -> NSUInteger,
     );
     class.add_method(sel!(draggingExited:), dragging_exited as extern "C" fn(&Object, Sel, id));
+    class.add_method(
+        sel!(handleNotification:),
+        handle_notification as extern "C" fn(&Object, Sel, id),
+    );
 
     add_mouse_button_class_method!(class, mouseDown, ButtonPressed, MouseButton::Left);
     add_mouse_button_class_method!(class, mouseUp, ButtonReleased, MouseButton::Left);
@@ -219,7 +238,7 @@ extern "C" fn accepts_first_mouse(_this: &Object, _sel: Sel, _event: id) -> BOOL
 extern "C" fn become_first_responder(this: &Object, _sel: Sel) -> BOOL {
     let state = unsafe { WindowState::from_view(this) };
     let is_key_window = unsafe {
-        let window: id = msg_send![state.window_inner.ns_view, window];
+        let window: id = msg_send![this, window];
         let is_key_window: BOOL = msg_send![window, isKeyWindow];
         is_key_window == YES
     };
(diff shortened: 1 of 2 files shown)


maxjvh:
there was a slight oversight where the focus events were not sent correctly if the window contained other things besides just the NSView that houses the plugin. should be fixed now though, also added some comments for clarity.

src/macos/window.rs
```diff
@@ -34,6 +39,7 @@ use crate::gl::{GlConfig, GlContext};
 
 pub struct WindowHandle {
     state: Rc<WindowState>,
+    _observer: NSWindowNotificationObserver,
```
glowcoil, line 42: I don't think the `WindowHandle` is really the right place to store this. If the `WindowHandle` is dropped without closing the window, the window will stay open but we'll stop receiving focus events; and if the window is closed without dropping the `WindowHandle`, the observer will stay registered and we'll continue trying to deliver focus events (which I think will result in some use-after-frees in `handle_notification`).

src/macos/window.rs
```diff
@@ -415,3 +424,90 @@ pub fn copy_to_clipboard(string: &str) {
...
+
+impl NSWindowNotificationObserver {
+    const NS_WINDOW_DID_BECOME_KEY_NOTIFICATION: &'static str = "NSWindowDidBecomeKeyNotification";
+    const NS_WINDOW_DID_RESIGN_KEY_NOTIFICATION: &'static str = "NSWindowDidResignKeyNotification";
+
+    fn new(window_state: Rc<WindowState>) -> Self {
+        unsafe {
+            // Create unique observer class and instantiate it.
+            let observer: id = {
+                let name = format!("BaseviewObserver_{}", Uuid::new_v4().to_simple());
```
glowcoil, line 467: This will leak a new class definition every time a new window is opened. To avoid this for our `NSView` subclass, we deregister the class definition in the [`dealloc` override](https://github.com/RustAudio/baseview/blob/26b019d6a278ac94d6011cebbfa13c8c9017306b/src/macos/view.rs#L229).

In fact, I think it should be possible to define the `handleNotification:` method directly on the `NSView` subclass and use that as the observer, similarly to how we use it as the `NSWindowDelegate` in [standalone mode](https://github.com/RustAudio/baseview/blob/26b019d6a278ac94d6011cebbfa13c8c9017306b/src/macos/window.rs#L241).

src/macos/window.rs
```diff
@@ -415,3 +424,90 @@ pub fn copy_to_clipboard(string: &str) {
...
+                    sel!(handleNotification:),
+                    handle_notification as extern "C" fn(&Object, Sel, id),
+                );
+                decl.add_ivar::<*mut c_void>(BASEVIEW_STATE_IVAR);
+                let class = decl.register();
+                msg_send![class, new]
+            };
+
+            // Attach a pointer to WindowState to the observer instance.
+            let window_state_ptr = Rc::into_raw(Rc::clone(&window_state));
```
glowcoil, line 479: This call to `Rc::into_raw` has no matching call to `Rc::from_raw`, which means that the `WindowState` will never get freed. You could fix this by adding cleanup code, but I think the suggestion above to use the `NSView` itself as the observer would also fix it in a simpler way.

glowcoil (changes requested):

src/macos/window.rs
maxjvh: using `NSView` as the observer is a great idea, thanks!

src/macos/view.rs
```diff
@@ -94,6 +94,18 @@ macro_rules! add_simple_keyboard_class_method {
...
 }
 
+unsafe fn register_notification(observer: id, notification_name: &'static str, object: id) {
+    let notification_center: id = msg_send![class!(NSNotificationCenter), defaultCenter];
+
+    let _: () = msg_send![
+        notification_center,
+        addObserver:observer
+        selector:sel!(handleNotification:)
+        name:NSString::alloc(nil).init_str(notification_name)
```
maxjvh, line 104: this NSString gets automatically released i assume?

src/macos/view.rs
```diff
@@ -225,6 +271,9 @@ extern "C" fn dealloc(this: &mut Object, _sel: Sel) {
         let superclass = msg_send![this, superclass];
         let () = msg_send![super(this, superclass), dealloc];
 
+        let notification_center: id = msg_send![class!(NSNotificationCenter), defaultCenter];
+        let () = msg_send![notification_center, removeObserver:this];
```
maxjvh, line 275: would feel better to remove the observer before calling dealloc on the superclass, but apparently the call moves `this` and couldn't figure out how to do it otherwise

src/macos/view.rs
```diff
@@ -473,3 +522,30 @@ extern "C" fn dragging_exited(this: &Object, _sel: Sel, _sender: id) {
...
+        // The NSWindow object associated with our NSView.
+        let window: id = msg_send![this, window];
+
+        let first_responder: id = msg_send![window, firstResponder];
+
+        // Only trigger focus events if the NSWindow that's being notified about is our window,
+        // and if the window's first responder is our NSView.
+        // If the first responder isn't our NSView, the focus events will instead be triggered
+        // by the becomeFirstResponder and resignFirstResponder methods on the NSView itself.
+        if notification_object == window && first_responder == state.window_inner.ns_view {
```
maxjvh, line 542: how to compare `first_responder: id` with `this: &Object`? would make things even more clean and make the only change in window.rs unnecessary

maxjvh (commented):
thanks for the review @glowcoil! making NSView the observer seemed to address all your comments. just a few questions here still.

src/macos/view.rs
glowcoil: I think the solution to this immediate problem would be to explicitly reborrow `this`, i.e. to pass `&*this` instead of `this` to `removeObserver:`. That should let you move this before the `[super dealloc]`.

However, I think it might be a better idea to just move this call out of `dealloc` and into `WindowInner::close`.

src/macos/view.rs
glowcoil: I don't think it does, but you should be able to just call `autorelease` on it.

I think a nicer solution would be to statically link in the necessary string constants (warning, I haven't tested this):

```rust
#[link(name = "AppKit", kind = "framework")]
extern "C" {
    static NSWindowDidBecomeKeyNotification: id;
    static NSWindowDidResignKeyNotification: id;
}```

src/macos/view.rs
glowcoil: It should just work to cast `this as id`.

src/macos/view.rs
maxjvh: this does work indeed! :)

src/macos/view.rs
maxjvh: worked when going through a const pointer first `this as *const Object as id`

glowcoil (approved):

PR #172, opened 2024-03-23
Add simple open_parented example

prokopyl:
This PR adds a simple example that allows to test and showcase the `Window::open_parented` method.

That example first creates a parent window using `Window::open_blocking`, and then creates a smaller child window using `Window::open_parented`.

Both window's handlers log all of their events to the console, in a similar fashion to the `open_window` example.

Both windows actually do rendering (unlike the `open_window` example for now): the parent fills its window with a grey backround, and the child fills its window with a red background.

This example also uses the `softbuffer` crate to perform the rendering, which allows testing it in a more portable manner and in the simplest use case possible, without having to involve OpenGL or any 3D rendering pipeline at all.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,4 @@ uuid = { version = "0.8", features = ["v4"] }
 
 [dev-dependencies]
 rtrb = "0.2"
+softbuffer = "0.3.4"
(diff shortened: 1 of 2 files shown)


glowcoil (approved):
Looks good to me! This will definitely be useful for testing going forward.

PR #174, opened 2024-03-24
Change `Window` from a borrowed handle to a weakly owned handle

prokopyl:
This PR implements the idea discussed with @glowcoil on Discord a few days ago, which is to make the `Window` handle `'static` and clone-able, so that it can be easily stored in the `WindowHandler`.

~~Note this PR is based on #172 for now, which is why the new example shows up in the diff. This can be changed depending on #172's merge status.~~ This has been properly rebased now.

This makes the following breaking API changes, for all platforms and backends:

* Removes the lifetime of the `Window` type.
* The `Window` type is now an owned, weak handle to the actual underlying window.
* Makes the `Window` type clone-able.
* The `Window` argument passed to the `WindowHandler` builder is now a fully owned value instead of a temporary borrow, which enables the handler to store it for its whole lifetime.
* Removes the `Window` argument on `WindowHandler::on_frame` and `WindowHandler::on_event` methods.
* All `Window`  methods now only take `&self` instead of `&mut self`.
* All `Window` methods are now internally fallible: they will panic if the actual window the handle points to has been destroyed. This can only happen if the `WindowHandler` misbehaves and clones the `Window` outside of its own lifetime.
* The `GlContext` has received the same treatment, for the same reasons: it's now a weak, owned clone-able handle to the GL context of a window, and its methods will also panic if the underlying context is destroyed.

As of now, this PR is still a WIP:

- [x] ~~The X11 backend is complete and functional, but it will probably need to be rebased on top of #173 before this PR can be merged.~~ It's done!
- [x] ~~I started working on the Win32 backend, but it's not done yet.~~ Done!
- [ ] ~~I haven't touched the Cocoa backend, and it will not build at all at the moment.~~ Done, but still needs testing.

As of now, this PR will not build on macOS or Windows until I've completed the work mentioned above.

### DIFF
--- a/examples/open_parented.rs
+++ b/examples/open_parented.rs
@@ -14,9 +14,9 @@ struct ParentWindowHandler {
 }
 
 impl ParentWindowHandler {
-    pub fn new(window: &mut Window) -> Self {
-        let ctx = unsafe { softbuffer::Context::new(window) }.unwrap();
-        let mut surface = unsafe { softbuffer::Surface::new(&ctx, window) }.unwrap();
+    pub fn new(window: Window) -> Self {
+        let ctx = unsafe { softbuffer::Context::new(&window) }.unwrap();
+        let mut surface = unsafe { softbuffer::Surface::new(&ctx, &window) }.unwrap();
         surface.resize(NonZeroU32::new(512).unwrap(), NonZeroU32::new(512).unwrap()).unwrap();
 
         let window_open_options = baseview::WindowOpenOptions {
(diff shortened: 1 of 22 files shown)


prokopyl:
Win32 backend update done, tested and functional! :slightly_smiling_face:

timboudreau:
I'd second this - I'm attempting to [integrate baseview as a replacement for winit in floem](https://github.com/lapce/floem/issues/505) and the lifetime (not to mention the requirement that the window creation closure be `Send`) and borrowed nature of it has caused me to have to jump through a lot of potentially unsound hoops just to get things painting (floem assumes it can hold a reference to *something* representing the window, and call it at times of its choosing on the event thread).

This would *vastly* simplify that integration.

prokopyl:
This has been discussed elsewhere, but it is pretty clear this PR is way too big and touches way too many things at once to be merged in its current state.

I will close this in favor of future, smaller changes that will eventually amount to this (most likely across several baseview releases)

PR #175, opened 2024-03-25
Render a background for the open_window example

prokopyl:
This PR adds code to render a basic gray background to the opened window in the `open_window` example.

~~This is done mostly because X11 can get fairly inconsistent with its order of operations when nothing is rendered to it, which can cause it to just never actually show the window.~~ Nevermind, that was a missing flush call on my end. :sweat_smile: 

This also helps making the example a bit more useful, since most users will want to render to their window.

And also it looks nicer. :slightly_smiling_face: 

This is done using the `softbuffer` crate, in the same manner of the `open_parented` introduced in #172.

### DIFF
--- a/examples/open_window.rs
+++ b/examples/open_window.rs
@@ -1,10 +1,13 @@
+use std::num::NonZeroU32;
 use std::time::Duration;
 
 use rtrb::{Consumer, RingBuffer};
 
 #[cfg(target_os = "macos")]
 use baseview::copy_to_clipboard;
-use baseview::{Event, EventStatus, MouseEvent, Window, WindowHandler, WindowScalePolicy};
+use baseview::{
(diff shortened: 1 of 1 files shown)


examples/open_window.rs
```diff
@@ -29,12 +44,22 @@ impl WindowHandler for OpenWindowExample {
...
                 #[cfg(target_os = "macos")]
                 match e {
-                    MouseEvent::ButtonPressed { .. } => {
-                        copy_to_clipboard(&"This is a test!")
-                    }
+                    MouseEvent::ButtonPressed { .. } => copy_to_clipboard(&"This is a test!"),
                     _ => (),
                 }
             }
+            Event::Window(WindowEvent::Resized(info)) => {
```
glowcoil, line 51: Nit: Move this inside the `Event::Window` case below?

glowcoil (approved):

PR #177, opened 2024-03-26
X11: split off Visual Info negotiation into a separate module

prokopyl:
This PR refactors and splits off the X11 visual info negotiation code into its own module and type. This should not change the actual logic at all however.

This is part of the effort to split up #174 into smaller pieces.

### DIFF
--- a/src/x11/visual_info.rs
+++ b/src/x11/visual_info.rs
@@ -12,7 +12,7 @@ pub(super) struct WindowVisualConfig {
 
     pub visual_depth: u8,
     pub visual_id: Visualid,
-    pub is_copy_from_parent: bool,
+    pub color_map: Option<Colormap>,
 }
 
 // TODO: make visual negotiation actually check all of a visual's parameters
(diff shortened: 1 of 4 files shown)


src/x11/visual_info.rs
```diff
@@ -0,0 +1,98 @@
...
+            fb_config: None,
+            visual_depth: COPY_FROM_PARENT as u8,
+            visual_id: COPY_FROM_PARENT,
+            is_copy_from_parent: true,
+        }
+    }
+
+    // For this 32-bit depth to work, you also need to define a color map and set a border
+    // pixel: https://cgit.freedesktop.org/xorg/xserver/tree/dix/window.c#n818
+    pub fn create_color_map(
```
glowcoil, line 65: In my opinion it would be cleaner for this to happen directly inside `find_best_visual_config`(`_for_gl`) rather than indirectly passing along the `is_copy_from_parent` flag, but I'll leave it up to you.

glowcoil (approved):

src/x11/visual_info.rs
prokopyl: Ah yes you're right, I didn't realize I didn't need that `is_copy_from_parent` flag.
I've pushed a version that fixes that, if you wanna take another look before merging. :slightly_smiling_face:

glowcoil: Looks good!

PR #178, opened 2024-03-27
Win32: Refactor, split off and fix resize logic and DPI changes handling

prokopyl:
This PR refactors and splits off the logic that handles re-sizing and discovering/updating the window's DPI.

Some of that logic is split off into a new, low-level `Win32Window` type, which only handles translating baseview types and assumptions into `Win32` calls, separately from managing events or the window handler.

This is part of the effort to split up #174 into smaller pieces.

This PR actually changes the logic, mainly by also splitting the internal `window_info` into two separate components, `current_size` (physical) and `current_scale_factor`, which are instead updated independently during their respective events.

Incidentally, this PR also fixes a few bugs:
* Fix the window handler not being notified of the new scaling factor that's discovered *right after* the window is created.
* Fix a `WM_DPICHANGE` event triggering a window resize but not triggering a resize event in the window handler. This supersedes the work done in #140.
* Fix handling of the `WM_DPICHANGE` event to use the OS-provided window dimensions *and* position, instead of computing our own dimensions and leaving the position as-is. This avoids a DPI-change loop due to the window constantly switching between two monitors as it's being dragged across them using the mouse.

This has all been tested on my Windows 10 machine and works as expected.

### DIFF
--- a/src/win/mod.rs
+++ b/src/win/mod.rs
@@ -1,6 +1,7 @@
 mod cursor;
 mod drop_target;
 mod keyboard;
+mod win32_window;
 mod window;
 
 pub use window::*;
--- a/src/win/win32_window.rs
+++ b/src/win/win32_window.rs
@@ -1,0 +1,122 @@
+use crate::PhySize;
+use std::convert::TryFrom;
+use std::ffi::OsStr;
+use std::os::windows::ffi::OsStrExt;
+use std::ptr::null_mut;
+use winapi::shared::minwindef::{ATOM, DWORD};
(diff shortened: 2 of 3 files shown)


prokopyl:
While this is better than #174 was, this is still a bit too large as it has quite a bit of internal refactorings as well. I'll put this back to a draft while I work on a "lite", bugfix-only version of this.

prokopyl:
Closing this now, as this has been incorporated piecemeal in #235 , #259 and #261

PR #179, opened 2024-03-27
Fix temporary names left over from x11rb migration

glowcoil:
The `conn2` and `atoms2` names were meant to be temporary while incrementally moving the X11 backend from the `xcb` crate to the `x11rb` crate, but were accidentally left in.


prokopyl (approved):
Ah, yeah I wondered what the intention was but didn't give it a second thought. :sweat_smile: 

This little cleanup looks good to me!

PR #180, opened 2024-03-27
Fix warnings in examples

glowcoil:


prokopyl (approved):
Ah, yes I missed those, good catch!

PR #181, opened 2024-03-28
Fix GL X11 error handling

prokopyl:
This PR fixes a soundness issue inside the `XErrorHandler` utility function, which could use an xlib `Display` pointer that had no guarantee to be still valid.

This could happen because the underlying `XErrorEvent` was stored directly inside the returned error type, and the `Display` pointer it contained was only called to extract the error message during in `Debug` implementation, which could happen long after the associated `Display` had been destroyed.

This PR fixes this by extracting the error message upfront and storing it as a string as soon as the error happens.
This PR also fixes the `handle` method that was not properly marked `unsafe`.

### DIFF
--- a/src/gl/x11.rs
+++ b/src/gl/x11.rs
@@ -229,12 +229,12 @@ impl GlContext {
     }
 
     pub fn swap_buffers(&self) {
-        errors::XErrorHandler::handle(self.display, |error_handler| {
-            unsafe {
+        unsafe {
+            errors::XErrorHandler::handle(self.display, |error_handler| {
                 glx::glXSwapBuffers(self.display, self.window);
-            }
-            error_handler.check().unwrap();
-        })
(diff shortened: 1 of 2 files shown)


glowcoil (approved):
I don't 100% love the idea of eagerly fetching the error text, but I guess it's preferable to alternatives like having the error enum keep the display alive with reference counting.

prokopyl:
> I don't 100% love the idea of eagerly fetching the error text, but I guess it's preferable to alternatives like having the error enum keep the display alive with reference counting.

Yeah it didn't feel right to me either initially, but the X docs support that calling `XGetErrorText` within the error handler is safe (it doesn't do any server requests). Plus it's just formatting a few strings which is negligible performance-wise, even more so considering it's only for the cold (error) path on the 3 functions we're calling.

PR #182, opened 2024-03-28
Add additional CI checks

prokopyl:
This PR adds a variety of checks to the CI:

* Check that all code and examples build without warnings using `-D warnings`. This is done for all platforms, with or without the OpenGL feature.
* Check that all documentation is valid, and that no documentation links are broken, using `cargo doc` with `-D warnings`.
* Run `clippy` with `-D warnings` (but otherwise only the default lints).
* Check formatting using `rustfmt`.

This PR also applies some light formatting and code fixes that were detected by `rustfmt` and `clippy`.

This also makes some updates to the actions used:
* Update `actions/checkout` to `v4`
* Replace `actions-rs/toolchain` (which is not maintained anymore) to `dtolnay/rust-toolchain` instead

Some further work that is not covered by this PR:
* The `missing-safety-doc` lint has been temporarily disabled. This will be re-enabled in another PR that will add the proper safety docs to the affected functions.
* `cargo-deny` has been discussed, but I prefer adding it in a later PR as it can be its own can of worms regarding licenses and dependencies.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -6,7 +6,6 @@ env:
 
 jobs:
   build:
-    name: Build & Test
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
@@ -34,7 +33,7 @@ jobs:
       run: cargo doc --examples --all-features --no-deps
     - name: Clippy
       run: cargo clippy --workspace --all-targets --all-features -- -D warnings
-    - name: Format
+    - name: Check Formatting (rustfmt)
       run: cargo fmt --all -- --check
 
 
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -5,7 +5,7 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  test:
+  build:
     name: Build & Test
     strategy:
       matrix:
@@ -23,6 +23,7 @@ jobs:
       uses: dtolnay/rust-toolchain@stable
       with:
         toolchain: stable
+        components: rustfmt, clippy
     - name: Build Default
       run: cargo build --workspace --all-targets --verbose
     - name: Build All Features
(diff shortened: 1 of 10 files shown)


.github/workflows/rust.yml
```diff
@@ -1,29 +1,62 @@
...
+      uses: dtolnay/rust-toolchain@stable
       with:
         toolchain: stable
-        override: true
-    - name: Build with default features
-      run: cargo build --examples --workspace --verbose
-    - name: Build again with all features
-      run: cargo build --examples --workspace --all-features --verbose
+    - name: Build Default
+      run: cargo build --all --verbose
```
glowcoil, line 27: I think this should be `cargo build --workspace --all-targets`. `--all` is a deprecated alias for `--workspace` (see [here](https://doc.rust-lang.org/cargo/commands/cargo-build.html#package-selection)), which only specifies affects package selection, not target selection, so without `--examples` or `--all-targets`, the examples will no longer be built.

glowcoil (changes requested):

.github/workflows/rust.yml
prokopyl: Ah, my bad, I thought it was the opposite. :sweat_smile:

prokopyl: Fixed now :slightly_smiling_face:

.github/workflows/rust.yml
```diff
@@ -1,29 +1,62 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: stable
+          components: clippy
+      - name: Install XCB and GL dependencies
+        if: contains(matrix.os, 'ubuntu')
+        run: sudo apt-get install libx11-xcb-dev libxcb-dri2-0-dev libgl1-mesa-dev libxcb-icccm4-dev libxcursor-dev
+      - run: cargo clippy --all --all-features -- -D warnings
```
glowcoil, line 50: I think `cargo clippy` will also take `--workspace --all-targets` :-)

glowcoil (changes requested):

.github/workflows/rust.yml
prokopyl: Dangit lol, fixed now x)

glowcoil (approved):

PR #184, opened 2024-03-28
Call XCloseDisplay instead of xcb_disconnect

glowcoil:
Pass `false` for the `should_drop` argument of [`XCBConnection::from_raw_xcb_connection`](https://docs.rs/x11rb/0.13.0/x11rb/xcb_ffi/struct.XCBConnection.html#method.from_raw_xcb_connection), so that it doesn't implicitly call `xcb_disconnect` on drop. Instead, explicitly call `XCloseDisplay` on drop, since the connection was originally created with `XOpenDisplay`.

### DIFF
--- a/src/x11/xcb_connection.rs
+++ b/src/x11/xcb_connection.rs
@@ -40,7 +40,7 @@ impl XcbConnection {
         let xcb_connection = unsafe { xlib_xcb::XGetXCBConnection(dpy) };
         assert!(!xcb_connection.is_null());
         let screen = unsafe { xlib::XDefaultScreen(dpy) } as usize;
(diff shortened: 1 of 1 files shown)


prokopyl (approved):
The fix looks good to me, although I'm not sure why we're even dealing with `XOpenDisplay` / `XCloseDisplay` instead of using `XCBConnection::connect()`?

glowcoil:
> I'm not sure why we're even dealing with `XOpenDisplay` / `XCloseDisplay` instead of using `XCBConnection::connect()`?

It's necessary since we use the Xlib GLX API for OpenGL context creation. If we were to switch from GLX to EGL, it should be possible to stop linking to Xlib and xcb entirely and just use x11rb's `RustConnection`.

PR #185, opened 2024-03-30
Remove unnecessary dependences from CI pipeline and README

glowcoil:
After switching to `x11rb`, the `libxcb-icccm4-dev` and `libxcursor-dev` dependencies are no longer necessary. I don't believe the `libxcb-dri2-0-dev` dependency was ever necessary.


prokopyl (approved):
Tested on my end, looks good to me!

PR #186, opened 2024-04-01
Handle set_mouse_cursor event on windows

Fredemus:
This adds the ability to change the cursor on Windows platforms. For some reason there are very few default cursors included in Windows so a lot of cursors available on other platforms aren't available yet, which is why many of the `MouseCursor` options just maps to `IDC_ARROW`. I'll look into adding custom cursors next.

`LoadCursorW` is supposedly superseded by [LoadImageW](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadcursorw) but I couldn't find a way to do it with that one that doesn't crash. 

Tested with Vizia's cursor icon example [here](https://github.com/Fredemus/vizia/tree/baseview-window-events) and everything seems to work as it should.
Also tested by hovering over some buttons on a plugin.

### DIFF
--- a/src/win/cursor.rs
+++ b/src/win/cursor.rs
@@ -1,0 +1,55 @@
+use crate::MouseCursor;
+use winapi::{
+    shared::ntdef::LPCWSTR,
+    um::winuser::{
+        IDC_APPSTARTING, IDC_ARROW, IDC_CROSS, IDC_HAND, IDC_HELP, IDC_IBEAM, IDC_NO, IDC_SIZEALL,
+        IDC_SIZENESW, IDC_SIZENS, IDC_SIZENWSE, IDC_SIZEWE, IDC_WAIT,
+    },
+};
+
+pub fn cursor_to_lpcwstr(cursor: MouseCursor) -> LPCWSTR {
+    match cursor {
+        MouseCursor::Default => IDC_ARROW,
+        MouseCursor::Hand => IDC_HAND,
+        MouseCursor::HandGrabbing => IDC_SIZEALL,
+        MouseCursor::Help => IDC_HELP,
+
+        // hiding cursor can't be turned into an lpcwstr since it's done by ShowCursor instead of SetCursor
+        MouseCursor::Hidden => IDC_ARROW,
+
+        MouseCursor::Text => IDC_IBEAM,
+        MouseCursor::VerticalText => IDC_IBEAM,
+
+        MouseCursor::Working => IDC_WAIT,
+        MouseCursor::PtrWorking => IDC_APPSTARTING,
+
+        MouseCursor::NotAllowed => IDC_NO,
+        MouseCursor::PtrNotAllowed => IDC_NO,
+
+        MouseCursor::ZoomIn => IDC_ARROW,
+        MouseCursor::ZoomOut => IDC_ARROW,
+
+        MouseCursor::Alias => IDC_ARROW,
+        MouseCursor::Copy => IDC_ARROW,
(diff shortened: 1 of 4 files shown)


src/win/window.rs
```diff
@@ -428,6 +430,24 @@ unsafe fn wnd_proc_inner(
...
+            let low_word = (lparam & 0xFFFF) as i16 as isize;
+            let mouse_in_window = low_word == HTCLIENT;
+            if mouse_in_window {
+                // Here we need to set the cursor back to what the state says, since it can have changed when outside the window
+                let cursor =
+                    LoadCursorW(null_mut(), cursor_to_lpcwstr(*window_state.cursor_icon.borrow()));
+                unsafe {
+                    SetCursor(cursor);
+                }
+                Some(0)
```
glowcoil, line 445: I think this should return 1, not 0, though from what I can tell it only makes a difference in cases where the `WM_SETCURSOR` message originates from a child window of the Baseview window. More details [here](https://devblogs.microsoft.com/oldnewthing/20061121-15/?p=28943).

src/win/window.rs
```diff
@@ -790,8 +821,9 @@ impl Window<'_> {
         self.state.deferred_tasks.borrow_mut().push_back(task);
     }
 
-    pub fn set_mouse_cursor(&mut self, _mouse_cursor: MouseCursor) {
-        todo!()
+    pub fn set_mouse_cursor(&mut self, mouse_cursor: MouseCursor) {
+        let task = WindowTask::SetMouseCursor(mouse_cursor);
```
glowcoil, line 825: Why does this need to be a `WindowTask`? It should be fine for `set_mouse_cursor` to set the cursor directly.

glowcoil (changes requested):

src/win/window.rs
```diff
@@ -480,6 +500,7 @@ pub(super) struct WindowState {
     keyboard_state: RefCell<KeyboardState>,
     mouse_button_counter: Cell<usize>,
     mouse_was_outside_window: RefCell<bool>,
+    cursor_icon: RefCell<MouseCursor>,
```
glowcoil, line 503: This can just be a `Cell` instead of a `RefCell`.

Fredemus (commented):
Thanks for the quick review, I appreciate it! Newest commit should resolve all your comments

src/win/cursor.rs
```diff
@@ -0,0 +1,55 @@
...
+
+pub fn cursor_to_lpcwstr(cursor: MouseCursor) -> LPCWSTR {
+    match cursor {
+        MouseCursor::Default => IDC_ARROW,
+        MouseCursor::Hand => IDC_HAND,
+        MouseCursor::HandGrabbing => IDC_SIZEALL,
+        MouseCursor::Help => IDC_HELP,
+
+        // hiding cursor can't be turned into an lpcwstr since it's done by ShowCursor instead of SetCursor
+        MouseCursor::Hidden => IDC_ARROW,
```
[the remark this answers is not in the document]
glowcoil: It's just an alternative API that works slightly differently. I think all of Windows, macOS, and X11 have both a "show/hide cursor" API and a way to set the cursor to an empty image. The latter is just a closer fit for Baseview's API, since it treats "hidden" as one of the possible cursor states rather than having separate methods for showing and hiding the cursor.

src/win/window.rs
glowcoil: Yeah, `None` means `return DefWindowProc(...)` and `Some(x)` means `return x`.

PR #191, opened 2024-05-12
simple wgpu example

andrew-r-thomas:
this is a simple example using wgpu of a rainbow triangle that is centered at the mouse position.

it's probably not the cleanest way to do things on the wgpu side but it's a good starter example for people looking to use wgpu in baseview

some things to note:
- the wgpu version used is 0.18, this is because we needed a version that accepted a window that implements `HasRawWindowHandle` and `HasRawDisplayHandle` from rwh
- the surface size on wgpu should be twice the window size in baseview for the sizes to match up. not sure why this is, but good to note

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ rtrb = "0.2"
 softbuffer = "0.3.4"
 
 [workspace]
-members = ["examples/render_femtovg"]
+members = ["examples/render_femtovg", "examples/render_wgpu"]
 
 [lints.clippy]
 missing-safety-doc = "allow"
--- a/examples/render_wgpu/Cargo.toml
+++ b/examples/render_wgpu/Cargo.toml
@@ -7,5 +7,5 @@ edition = "2021"
 
 [dependencies]
 wgpu = "0.18"
-baseview = {path = "../.."}
+baseview = {path = "../..", features = ["opengl"]}
 pollster = "0.3.0"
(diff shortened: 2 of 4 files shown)


prokopyl (changes requested):
Thanks for this PR! :slightly_smiling_face: 

The issue of the surface size needing to be twice as big as the window size is a window scaling factor issue. You probably have a High-DPI monitor on your machine, but on my machine everything is off by a factor of 2.

To fix this you need to handle window resize events, which also carry the scaling factor. Currently in `baseview` there is no way to get the actual window size at creation time, but a resize event will be fired right away which will give you all the information you need.

You can use the `render_femtovg` example as a reference on how to handle those events and resize your rendering buffers when needed. :slightly_smiling_face: 

You're also recreating the whole vertex buffer on every mouse move, but I think it would be better to simply write the mouse position into a uniform buffer and handle the translation in the shader instead.

~~There's also a couple clippy warnings that need fixing, but nothing too bad.~~ Nevermind, seems like you fixed those already! :+1:

andrew-r-thomas:
@prokopyl not sure what the weird clippy error is for the windows test, looks like its coming from the baseview source. everything else should be fixed though!

murl-digital:
fyi @andrew-r-thomas, i was able to get the latest version of wgpu working with your example, basically i remove the lifetime requirements, made ``Surface`` ``'static``, and converted directly between the raw window and display handle types between 05 and 0.6

andrew-r-thomas:
> fyi @andrew-r-thomas, i was able to get the latest version of wgpu working with your example, basically i remove the lifetime requirements, made `Surface` `'static`, and converted directly between the raw window and display handle types between 05 and 0.6

sweet! I think there are a bunch of refactoring changes for baseview that are in progress that will allow for 0.6 to be more fully supported, but good to know that theres a fix in the meantime!

PR #194, opened 2024-07-12
x11: fix key modifiers mappings

Quant1um:
this pull request fixes a bug on x11 where ALT modifies would be triggered by holding left mouse button instead of pressing alt key (same for NUM_LOCK and META modifiers)


BillyDM:
I'm using this branch for my `egui_baseview` crate, and I can confirm that it fixes the issue.

micahrj (approved):

Issue #202, opened 2025-02-10
macOS/Linux focus issue


justinfrankel:
I'm one of the developers of REAPER. There is an issue where plug-is that use nih-plug and egui (which afaict uses baseview) do not set their NSViews as first responder on mouse click. On most other DAWs this is not an issue, as the NSView is often the only view in the NSWindow, however in REAPER it does present some issues.

I have written some suggested code that should fix for macOS, which compiles, but due to my unfamiliarity with the Cargo ecosystem I haven't been able to get it into the actual resulting binary.

`diff --git a/src/macos/view.rs b/src/macos/view.rs
index 8b765b4..a970b33 100644
--- a/src/macos/view.rs
+++ b/src/macos/view.rs
@@ -54,12 +54,18 @@ macro_rules! add_simple_mouse_class_method {
 /// Similar to [add_simple_mouse_class_method!], but this creates its own event object for the
 /// press/release event and adds the active modifier keys to that event.
 macro_rules! add_mouse_button_class_method {
-    ($class:ident, $sel:ident, $event_ty:ident, $button:expr) => {
+    ($class:ident, $sel:ident, $event_ty:ident, $button:expr, $foc:expr) => {
         #[allow(non_snake_case)]
         extern "C" fn $sel(this: &Object, _: Sel, event: id){
             let state = unsafe { WindowState::from_view(this) };
 
             let modifiers = unsafe { NSEvent::modifierFlags(event) };
+            if $foc {
+                unsafe {
+                    let window: id = msg_send![this, window];
+                    let _: () = msg_send![window, makeFirstResponder: this];
+                }
+            }
 
             state.trigger_event(Event::Mouse($event_ty {
                 button: $button,
@@ -211,12 +217,12 @@ unsafe fn create_view_class() -> &'static Class {
         handle_notification as extern "C" fn(&Object, Sel, id),
     );
 
-    add_mouse_button_class_method!(class, mouseDown, ButtonPressed, MouseButton::Left);
-    add_mouse_button_class_method!(class, mouseUp, ButtonReleased, MouseButton::Left);
-    add_mouse_button_class_method!(class, rightMouseDown, ButtonPressed, MouseButton::Right);
-    add_mouse_button_class_method!(class, rightMouseUp, ButtonReleased, MouseButton::Right);
-    add_mouse_button_class_method!(class, otherMouseDown, ButtonPressed, MouseButton::Middle);
[...]

PR #203, opened 2025-03-17
feat: add focus handling to macOS view for proper key event forwarding

andrewrynhard:
Updated the `add_mouse_button_class_method!` macro.
Now, when `$foc` is true, the code calls `makeFirstResponder` on the window to ensure the view becomes the first responder.
This change enables the plugin view to properly capture keyboard input on macOS.

Credit to @justinfrankel. See #202.

### DIFF
--- a/src/macos/view.rs
+++ b/src/macos/view.rs
@@ -54,13 +54,18 @@ macro_rules! add_simple_mouse_class_method {
 /// Similar to [add_simple_mouse_class_method!], but this creates its own event object for the
 /// press/release event and adds the active modifier keys to that event.
 macro_rules! add_mouse_button_class_method {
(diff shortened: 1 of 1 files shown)


andrewrynhard:
Fixes #191.

kronihias:
This fixes keyboard focus for plugins within Reaper for me (eg. a textfield).

Any chance we can get this merged?

micahrj:
This is unnecessary. There is already a `Window::focus` method which calls `makeFirstResponder` on macOS, which applications can call explicitly e.g. in response to mouse down events: https://github.com/RustAudio/baseview/blob/237d323c729f3aa99476ba3efa50129c5e86cad3/src/macos/window.rs#L302-L310

PR #204, opened 2025-04-06
avoid crash when window is null in become_first_responder

httnn:
apparently newer Rust versions (starting with 1.86?) have stricter debug assertions regarding unsafe code which caused a crash under debug builds because `window` was null in the `become_first_responder` method.

this fix simply checks whether `window` is null or not.


micahrj (approved):

micahrj:
It looks like the build is failing due to warnings in the `objc` crate. Those warnings have been addressed on the master branch but not on the latest version on crates.io: https://github.com/SSheldon/rust-objc/issues/125

Long-term I think the best thing to do would be to switch to the `objc2` crate, but my recommendation for now is to add the following to the top of lib.rs:

```rust
#![allow(unexpected_cfgs)]
```

PR #205, opened 2025-04-24
fix drag & drop coordinates on Windows

httnn:
the coordinates were previously in screen space, not in the coordinate space of the hwnd (= "client").


httnn:
before (coordinates are in screen space):
https://github.com/user-attachments/assets/516fdf1d-22e3-4aec-98ba-47313cafa9c5

after (coordinates are in window space):
https://github.com/user-attachments/assets/30c93de3-0341-424d-bc48-9be1c95ac204

micahrj (approved):

PR #210, opened 2025-10-01
Use the user's keyboard layout from the x11 system

### LINKED ISSUE #94: Keyboard layout not accounted for in keyboard events
Right now the keyboard events assume a standard QWERTY layout.

TristanCacqueray:
This change uses xkbcommon to decode the key utf8 value based on the user's current layout. The implementation is adapted from the x11rb's xkbcommon-example.

### DIFF
--- a/src/wrappers/xkbcommon.rs
+++ b/src/wrappers/xkbcommon.rs
@@ -8,28 +8,30 @@ pub struct XkbcommonState {
 }
 
 impl XkbcommonState {
-    pub fn new(xcb_connection: &crate::x11::XcbConnection) -> Self {
-        let xkb_common = xkbc::xkbcommon_handle();
-        let context =
-            unsafe { (xkb_common.xkb_context_new)(xkbc::xkb_context_flags::XKB_CONTEXT_NO_FLAGS) };
+    pub fn new(xcb_connection: &crate::x11::XcbConnection) -> Option<Self> {
+        xkbc::xkbcommon_option().and_then(|xkb_common| {
+            xkbc::x11::xkbcommon_x11_option().and_then(|xkb_x11| {
+                let context = unsafe {
+                    (xkb_common.xkb_context_new)(xkbc::xkb_context_flags::XKB_CONTEXT_NO_FLAGS)
+                };
 
-        let conn: *mut xkbc::x11::xcb_connection_t =
-            xcb_connection.conn.xcb_connection().get_raw_xcb_connection();
+                let conn: *mut xkbc::x11::xcb_connection_t =
+                    xcb_connection.conn.xcb_connection().get_raw_xcb_connection();
 
-        let xkb_x11 = xkbc::x11::xkbcommon_x11_handle();
-
-        let state = unsafe {
-            let device_id = (xkb_x11.xkb_x11_get_core_keyboard_device_id)(conn);
-            assert!(device_id >= 0);
-            let keymap = (xkb_x11.xkb_x11_keymap_new_from_device)(
-                context,
-                conn,
-                device_id,
-                xkbc::xkb_keymap_compile_flags::XKB_KEYMAP_COMPILE_NO_FLAGS,
-            );
-            (xkb_x11.xkb_x11_state_new_from_device)(keymap, conn, device_id)
-        };
-        XkbcommonState { state, xkb_common }
+                let state = unsafe {
+                    let device_id = (xkb_x11.xkb_x11_get_core_keyboard_device_id)(conn);
+                    assert!(device_id >= 0);
+                    let keymap = (xkb_x11.xkb_x11_keymap_new_from_device)(
+                        context,
+                        conn,
+                        device_id,
+                        xkbc::xkb_keymap_compile_flags::XKB_KEYMAP_COMPILE_NO_FLAGS,
+                    );
+                    (xkb_x11.xkb_x11_state_new_from_device)(keymap, conn, device_id)
+                };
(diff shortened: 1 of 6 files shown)


prokopyl (changes requested):
Hi! And thanks for this PR! 🙂 

I have yet to delve into the details of the XKB implementation, but one issue I see is that the `xkbcommon-rs` crate links directly against the `xkbcommon` library, which would mean hosts would fail to load a plugin altogether.

It's for this reason we have moved to using `x11-dl` for instance. I think this should use
`xkbcommon-dl` instead (which is also what `winit` uses).

There's also been quite a few changes to the X11 codebase since you first submitted this PR, so if you have the time to rebase and update it, it would be very helpful!

TristanCacqueray:
Hey, alas I'm not sure how to get the keymap with `xkbcommon-dl`. Previously we could use `get_core_keyboard_device_id`, but now it looks like we need a name (e.g. for `xkb_keymap_new`)?

prokopyl:
> Hey, alas I'm not sure how to get the keymap with `xkbcommon-dl`. Previously we could use `get_core_keyboard_device_id`, but now it looks like we need a name (e.g. for `xkb_keymap_new`)?

So it looks like `xkbcommon-dl` has a `x11` feature that [exposes `xkb_x11_get_core_keyboard_device_id`](https://github.com/rust-windowing/xkbcommon-dl/blob/master/src/x11.rs) & other XCB functions that we need. It's just that they are not shown on docs.rs (probably because they haven't configured the `x11` feature to show up there), but it is available there!

TristanCacqueray:
@prokopyl thanks, that makes sense now. So I updated the PR and it seems to work as expected: in the open_window example, using a french layout, pressing the keycode for `2` emit the proper `é` Character. To be honest I'm not super familiar with xkbcommon, and even less with these `-dl` variants, so I'd be happy to rework this change if needed.

Do we need a safe fallback in case `xkbcommon_option` is None, e.g. to handle a missing libxkbcommon-x11.so ?

TristanCacqueray:
I also confirm this works downstream, e.g. through https://codeberg.org/TristanCacqueray/egui-baseview/src/branch/xkbfix

prokopyl:
> @prokopyl thanks, that makes sense now. So I updated the PR and it seems to work as expected: in the open_window example, using a french layout, pressing the keycode for `2` emit the proper `é` Character. To be honest I'm not super familiar with xkbcommon, and even less with these `-dl` variants, so I'd be happy to rework this change if needed.
> 
> Do we need a safe fallback in case `xkbcommon_option` is None, e.g. to handle a missing libxkbcommon-x11.so ?

Sweet, thanks! I have tested it and it works pretty much as expected on my French keyboard. :slightly_smiling_face: 

I think a fallback is always good to have yes. `xkbcommon` is pretty much ubiquitous on destkop Linux OSes, but it's Linux so there's no guarantee.
To me the fallback could be as simple as just using the previous behavior (from before this PR). That way it would fall back to pretty much raw key-codes that only work for QWERTY, which is good enough in my opinion.

TristanCacqueray:
@prokopyl please let me know if this PR needs more change, I'm looking forward getting this integrated. Thanks!

src/wrappers/xkbcommon.rs
```diff
@@ -0,0 +1,75 @@
...
+pub(crate) type Keycode = xkbcommon_dl::xkb_keycode_t;
+/// A xkbcommon state object
+pub struct XkbcommonState {
+    state: *mut xkbc::xkb_state,
+    xkb_common: &'static xkbc::XkbCommon,
+}
+
+impl XkbcommonState {
+    pub fn new(xcb_connection: &crate::x11::XcbConnection) -> Option<Self> {
+        xkbc::xkbcommon_option().and_then(|xkb_common| {
```
prokopyl: Small nitpick: you can use the `?` operator here instead of `and_then` to make that function more readable. 🙂

src/wrappers/xkbcommon.rs
```diff
@@ -0,0 +1,75 @@
...
+                        conn,
+                        device_id,
+                        xkbc::xkb_keymap_compile_flags::XKB_KEYMAP_COMPILE_NO_FLAGS,
+                    );
+                    (xkb_x11.xkb_x11_state_new_from_device)(keymap, conn, device_id)
+                };
+                Some(XkbcommonState { state, xkb_common })
+            })
+        })
+    }
```
prokopyl, line 40: There should be a `Drop` implementation here that frees the `state` and `context`, right now they are leaked every time a new window is opened.

prokopyl (changes requested):
There's a couple of minor changes left, but other than that it looks good to me! Thank you! 😃

src/wrappers/xkbcommon.rs
TristanCacqueray: Ha finally, I remembered `?` only worked with Result, that's great!

src/wrappers/xkbcommon.rs
TristanCacqueray: Good point, I added xkb_state_unref to the drop impl for `state`, but I'm not sure what you mean by `context`. The xkb_common is a static ref, so it shouldn't leak.

prokopyl: By context I mean the return value of `xkb_context_new` ^^

prokopyl (approved):
This looks good to me now! Thank you *very* much for your work, it is much appreciated!

TristanCacqueray:
You are welcome, thank you for maintaining baseview!

PR #212, opened 2025-10-20
Implement keyboard event hook on Windows

estroBiologist:
This PR implements **keyboard event hooking** on Windows, a technique also used by JUCE to get around restrictions on keyboard input imposed by various hosts. By intercepting Win32 messages using `SetWindowsHookEx`, we can bypass the host and forward keyboard messages to the `baseview` window directly.

This resolves the longstanding issue with keyboard input in Ableton, and likely renders many existing hacks for this problem obsolete (though I haven't tested that part yet.)

The meat of the implementation is in the new `src/win/hook.rs` module, which registers a simple event hook that intercepts keyboard-related `WM_` events addressed to `baseview` windows, and forwards them into the relevant window's event handler directly, consuming the message in the process.

Couple notes:
- There's some pretty nasty code for detecting `baseview` windows and forwarding events to them in `hook.rs`. If anyone has any better ideas for how to achieve this, I'm all ears, but it does seem to work as is.
- I had to add the `dtor` crate to the `windows` dependencies to properly detach the hook on DLL exit. Hope that's okay!
- I tried to keep the code reasonably clean, but neither my Win32 nor my Unsafe Rust skills are particularly great, so I hope it's all up to snuff.

**The keyboard hook in action:**
yesssss

### DIFF
--- a/src/win/drop_target.rs
+++ b/src/win/drop_target.rs
@@ -27,6 +27,7 @@ use super::WindowState;
 // These function pointers have to be stored in a (const) variable before they can be transmuted
 // Transmuting is needed because winapi has a bug where the pt parameter has an incorrect
 // type `*const POINTL`
+#[allow(non_snake_case)]
 const DRAG_ENTER_PTR: unsafe extern "system" fn(
     this: *mut IDropTarget,
     pDataObj: *const IDataObject,
@@ -34,12 +35,14 @@ const DRAG_ENTER_PTR: unsafe extern "system" fn(
     pt: POINTL,
     pdwEffect: *mut DWORD,
 ) -> HRESULT = DropTarget::drag_enter;
+#[allow(non_snake_case)]
 const DRAG_OVER_PTR: unsafe extern "system" fn(
     this: *mut IDropTarget,
     grfKeyState: DWORD,
     pt: POINTL,
     pdwEffect: *mut DWORD,
 ) -> HRESULT = DropTarget::drag_over;
+#[allow(non_snake_case)]
 const DROP_PTR: unsafe extern "system" fn(
     this: *mut IDropTarget,
     pDataObj: *const IDataObject,
--- a/src/win/window.rs
+++ b/src/win/window.rs
@@ -525,19 +525,19 @@ pub(super) struct WindowState {
 }
 
 impl WindowState {
-    pub(super) fn create_window(&self) -> Window {
+    pub(super) fn create_window(&self) -> Window<'_> {
         Window { state: self }
     }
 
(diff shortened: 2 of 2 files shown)


src/win/hook.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    if n_code == HC_ACTION && wparam == PM_REMOVE as usize && offer_message_to_baseview(msg) {
+        *msg = MSG {
+            hwnd: ptr::null_mut(),
+            message: WM_USER,
+            wParam: 0,
+            lParam: 0,
+            time: 0,
+            pt: POINT { x: 0, y: 0 },
+        };
```
micahrj: Zeroing out the `MSG` struct like this should not be necessary, as the way we prevent the message from propagating is just by not calling `CallNextHookEx`. It also seems like it is explicitly disallowed by the [docs](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nc-winuser-hookproc) for the `HOOKPROC` callback:

> The hook procedure can examine the message; it cannot modify it.

src/win/hook.rs
```diff
@@ -0,0 +1,144 @@
...
+};
+
+use crate::win::wnd_proc;
+
+static HOOK: Mutex<Option<KeyboardHook>> = Mutex::new(None);
+
+// track all windows opened by this instance of baseview
+// we use an RwLock here since the vast majority of uses (event interceptions)
+// will only need to read from the HashSet
+static OPEN_WINDOWS: LazyLock<RwLock<HashSet<HWNDWrapper>>> = LazyLock::new(|| RwLock::default());
```
micahrj: I think we should combine these two pieces of state into a single struct protected by the same `RwLock`. I don't think we gain anything by having them separate, and using a single mutex removes the risk of race conditions in the init/deinit logic.

src/win/hook.rs
```diff
@@ -0,0 +1,143 @@
...
+    match msg.message {
+        WM_KEYDOWN | WM_SYSKEYDOWN | WM_KEYUP | WM_SYSKEYUP | WM_CHAR | WM_SYSCHAR => {}
+
+        _ => return false,
+    }
+
+    // check if this is one of our windows. if so, intercept it
+    let state = HOOK_STATE.read().unwrap();
+
+    if state.open_windows.contains(&HWNDWrapper(msg.hwnd)) {
```
micahrj: It just occurred to me that it could result in a deadlock if we're holding the `HOOK_STATE` lock for the duration of the `wnd_proc` call and it opens or closes a window. We should make sure to drop the lock guard first.

PR #217, opened 2025-12-20
Win: always link with opengl32.lib

damyanp:
winapi explicitly doesn't link with opengl32.lib on all supported Windows platforms (aarch64 and thumbv7a). Presumably this dates back to when these Windows platforms didn't support OpenGL. Since that's no longer the case, and baseview won't be able to link successfully if the opengl feature is enabled and opengl32.lib isn't linked, this change makes it explicitly link opengl32.lib.

### DIFF
--- a/src/gl/win.rs
+++ b/src/gl/win.rs
@@ -20,6 +20,12 @@ use winapi::um::winuser::{
 
 use super::{GlConfig, GlError, Profile};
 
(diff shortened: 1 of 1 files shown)


prokopyl:
I'll be closing this in favor of #218. :slightly_smiling_face:

PR #221, opened 2026-01-16
Fix Windows OpenGL: Fallback to non-sRGB when sRGB not supported

### LINKED ISSUE #220: baseview crashes when compile for Windows and running under Wine
Normally, I wouldn't report this issue here, and just report it to wine (which I have done https://bugs.winehq.org/show_bug.cgi?id=59256), but in doing some experiments, I've noticed some unusual behavior that might point to other problems or might be useful in other ways, so here's a 'bug report' for something unusual.

Here's a repo that reproduces the behavior I'm going to discuss: https://github.com/john-parton/egui-wine-test (Ignore 'egui' in the name. It previously had baseview_egui as a dep, but I've simplified it to just use baseview)

Essentially, when attempting to initialize OpenGL, the program crashes. I have found that disabling sRGB support resolve the crash. 

Here I would normally say this is "Wine and Windows being difficult", but I did the same thing use egui, eframe and glow (the gl rendered) and it worked just fine, even under Wine.

This is the result of me going down a loooong rabbit hole. I just wanted the Spectral Analyzer VST Plugin to work nicely in FL Studio without "tricks." https://github.com/robbert-vdh/nih-plug

john-parton:
Previously, if sRGB framebuffer was requested but not available on Windows, the pixel format selection would fail silently with num_formats=0, leading to undefined behavior with an invalid pixel format.

This commit adds:
- Proper validation that wglChoosePixelFormatARB found a matching format
- Graceful fallback: retry without sRGB if initial request fails
- Warning message to inform users when falling back to non-sRGB

### DIFF
--- a/src/gl/win.rs
+++ b/src/gl/win.rs
@@ -203,6 +203,7 @@ impl GlContext {
 
         let hdc = GetDC(hwnd);
 
+        // Try to choose pixel format with requested config
         #[rustfmt::skip]
         let pixel_format_attribs = [
             WGL_DRAW_TO_WINDOW_ARB, 1,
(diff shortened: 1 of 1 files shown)


john-parton:
This resolves https://github.com/RustAudio/baseview/issues/220

It's a relatively simple workaround. As far as I can tell, the spec doesn't really say what happens if you use `0` as your value there, but I think `0` is really `null`/`None` in this context, so double checking and finding an actual appropriate value makes sense to me.

I didn't check the behavior on MacOS or Linux, as I wanted to keep the scope of this low, but there may be a similar issue.

john-parton:
It looks like the behavior under linux to is to return as an error:

https://github.com/RustAudio/baseview/blob/237d323c729f3aa99476ba3efa50129c5e86cad3/src/gl/x11.rs#L190-L192

I think probably it would be better if Windows behaved the same and then I guess whatever downstream libraries from here would have to handle that gracefully by explicitly re-initializing.

john-parton:
My preference is that this pull request be seriously considered, but I have provided a minimal alternate of propagating the error to the user here: https://github.com/RustAudio/baseview/pull/222 I'm not sure that pull request could even be considered done, but it's there if we want to pivot.

prokopyl:
I think the ideal solution would be to chuck GL context creation out of Baseview and just use `glutin`, which has a way more robust and better-tested implementation.

However, that ideal solution requires a breaking change (and probably more discussion), so I'm putting that aside and looking at this only for bug-fixing in the context of releasing v0.1. :slightly_smiling_face: 

I have yet to still test this on my end, but I much prefer this fallback approach compared to the one you implemented in #222, for two reasons:

* As you explained, error detection and recovery would be very difficult for downstream crates, *especially* since we don't have proper error management at the moment and most things just panic.
* This is actually the approach used by `glutin` (i.e. silently dropping the value of the `srgb` flag in the config if not supported, at least on Windows). This should make baseview in line with the libraries that use glutin+winit (which I have verified to be the case for at least egui and Vizia)

I'll report back with my testing, and will close #222 in favor of this PR. :slightly_smiling_face:

prokopyl (approved):
Code wise this looks good to me! 🙂

prokopyl:
Can confirm this works as intended, both on Wine and on native Windows.

Thank you for this PR, as well for the in-depth investigation, it is much appreciated! :slightly_smiling_face:

PR #222, opened 2026-01-16
Fix Windows OpenGL: Fail immediately when pixel format not available

john-parton:
Previously, if the requested pixel format (including sRGB) was not available on Windows, wglChoosePixelFormatARB would return num_formats=0, but the code would proceed with an invalid pixel_format value of 0, leading to undefined behavior.

This commit adds proper validation to match Linux behavior:
- Check if wglChoosePixelFormatARB found a matching format (num_formats <= 0)
- Check if pixel_format is invalid (== 0)
- Return GlError::CreationFailed immediately if validation fails
- Properly clean up (ReleaseDC) before returning error
- Add debug message indicating the failure and whether sRGB was requested

This makes Windows behave consistently with the Linux/GLX implementation, which also fails immediately with InvalidFBConfig if the requested framebuffer configuration is not supported.

### DIFF
--- a/src/gl/win.rs
+++ b/src/gl/win.rs
@@ -233,6 +233,12 @@ impl GlContext {
             &mut num_formats,
         );
 
+        if num_formats <= 0 || pixel_format == 0 {
+            eprintln!("Error: Failed to find matching pixel format (sRGB requested: {})", config.srgb);
+            ReleaseDC(hwnd, hdc);
+            return Err(GlError::CreationFailed(()));
+        }
+
         let mut pfd: PIXELFORMATDESCRIPTOR = std::mem::zeroed();
         DescribePixelFormat(
             hdc,


john-parton:
This partially resolves: https://github.com/RustAudio/baseview/issues/220

It is an alternate to this pull request: https://github.com/RustAudio/baseview/pull/221

This change will likely require users of this library to explicitly recover from a failed initialization. With this super-minimal change, I'm not even sure it's possible in a reasonable way, because the caller will have to just infer it's the srgb setting, despite the error being generic. Most likely a specific error enum should be added if we want to support explicitly gracefully failing and provider the caller with the information they need to try again successfully, although that puts us very-much-on-the-way to "just always retry for the user".

I have a relatively strong preference for the other pull request that just automatically selects a reasonable pixel format, because it just fixes the behavior I'm dealing with while working in https://github.com/robbert-vdh/nih-plug

But I've made this available for discussing.

prokopyl:
Closing this in favor of #221 (see [this comment](https://github.com/RustAudio/baseview/pull/221#issuecomment-4363906951) for the rationale behind this).

PR #224, opened 2026-03-30
add default methods to WindowOpenOptions

BillyDM:
This will fix an annoying issue I'm having in my NIH-plug fork where rust-analyzer will get confused when some crates use the "opengl" feature and some don't. I have to add special crate features just to work around it.

With this, I can just use the `..Default::default()` pattern to make both rust-analyzer and the rust compiler happy.

### DIFF
--- a/src/window_open_options.rs
+++ b/src/window_open_options.rs
@@ -18,7 +18,7 @@ pub enum WindowScalePolicy {
 pub struct WindowOpenOptions {
     pub title: String,
 
-    /// The logical size of the window.
+    /// The logical size of the window
     ///
     /// These dimensions will be scaled by the scaling policy specified in `scale`. Mouse
     /// position will be passed back as logical coordinates.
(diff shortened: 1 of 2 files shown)


src/window_open_options.rs
```diff
@@ -24,6 +29,32 @@ pub struct WindowOpenOptions {
 
     /// If provided, then an OpenGL context will be created for this window. You'll be able to
     /// access this context through [crate::Window::gl_context].
+    ///
+    /// By default this is set to `Some(GlConfig::default())`.
```
micahrj: Why default to `Some(GlConfig::default())`? It makes a lot more sense to me for it to default to `None`.

BillyDM: If the user is using a GUI library that expects OpenGL, then the GUI would fail to open with the default settings.

Though I suppose the GUI library layer could add the OpenGL config itself. So I guess either method is fine.

micahrj: I don't think we should be trying to guess what GUI libraries expect. Some GUI libraries may be using wgpu or software rendering, in which case defaulting to creating an OpenGL context will also cause problems. `gl_config` should default to `None`, and either the user or the GUI library should explicitly set it to something else if needed.

BillyDM:
Alright, I made `gl_config` default to `None`.

prokopyl (approved):
I agree it makes more sense to not create an OpenGL context by default. With this change done this looks good to me!

Issue #229, opened 2026-04-24
macOS builds failing with "unexpected `cfg` condition value: `cargo-clippy`"


micahrj:
This is happening due to an issue in the `objc` crate: https://github.com/SSheldon/rust-objc/issues/125

I think that ideally Baseview would just switch to the `objc2` crate, but that would require a decent amount of work since it's not a drop-in replacement. The simpler workaround for now would probably be to point at the master branch of the `rust-objc` repo, since a fix for the issue has been merged but was never published to crates.io.

micahrj:
I gave this a try and ran into the issue that the `cocoa` crate also depends on the `objc` crate, so the workaround would require forking `cocoa` as well.

paravozz:
@micahrj
I dug into this earlier, tested a small `build.rs` workaround that silences the warning locally:

```rust
fn main() {
  println!("cargo:rustc-check-cfg=cfg(feature, values(\"cargo-clippy\"))");
}
```

Happy to help with either, or with the `objc2` migration.

**UPD:** just in case, to speed things up, filed a PR with that fix: #231.

**UPD2:** Filed #232 with the `objc2` macros-only migration; supersedes #231. Kept `cocoa` for framework type bindings to leave the bigger cocoa-replacement decision open, since you flagged that direction. PR also folds in fixes for several latent `@encode` and `NSException` traps the stricter objc2 verification surfaces.

micahrj:
It looks like we already had `#![allow(unexpected_cfgs)]` attributes in the relevant modules, but for some reason they have stopped working:
1. https://github.com/RustAudio/baseview/blob/237d323c729f3aa99476ba3efa50129c5e86cad3/src/macos/mod.rs#L3
2. https://github.com/RustAudio/baseview/blob/237d323c729f3aa99476ba3efa50129c5e86cad3/src/gl/macos.rs#L3

I've found that if I move the attribute to the crate root it successfully silences the warning again: #233

PR #232, opened 2026-04-25
macOS: migrate from objc to objc2

paravozz:
## Summary

- Replaces `objc 0.2.7` runtime types and macros with `objc2 0.6` across `src/macos/` and `src/gl/macos.rs`. Keeps `cocoa = "0.24"` for framework type bindings (NSEvent, NSView, NSWindow, etc.); pure macros migration, no public-API or behavioral change.
- Drops `#![allow(unexpected_cfgs)]` from both files (no longer needed: objc2's macros don't expand the buggy `cargo-clippy` cfg, so the warning that justified it goes silent at the source).
- Resolves the `unexpected_cfgs` lint failure that was making CI fail under `RUSTFLAGS=-D warnings`. Related to #229; this PR is one of two paths discussed there (macros-only vs. full cocoa replacement) and intentionally takes the smaller-scope macros-only route to leave the cocoa replacement decision open. Supersedes #231 (the `build.rs` workaround), same problem fixed at the source rather than masked.
- Migration also surfaces and fixes several latent bugs in `keyboard.rs` and `view.rs` that objc 0.2's lax verification accepted but Apple's actual `@encode` requires; details below.

## What changed

**Direct replacements (mechanical):**

- `objc::msg_send!` → `objc2::msg_send!` (~54 sites across `view.rs`, `window.rs`, `keyboard.rs`, `gl/macos.rs`)
- `objc::class!` → `objc2::class!` (3 sites)
- `objc::sel!` → `objc2::sel!` (26 sites)
- `objc::declare::ClassDecl` → `objc2::runtime::ClassBuilder`
- `objc::runtime::Object` / `Class` / `Sel` / `Protocol` → `objc2::runtime::AnyObject` / `AnyClass` / `Sel` / `AnyProtocol`
- `objc::{Encode, Encoding}` → `objc2::{Encode, Encoding}`
- `objc::runtime::objc_disposeClassPair` → `objc2::ffi::objc_disposeClassPair`

**Less-mechanical patterns required by objc2's stricter API:**

- Callback signatures use `*const AnyObject` / `*mut AnyObject` instead of `&AnyObject` / `&mut AnyObject`. objc2 0.6's `MethodImplementation` trait impls aren't HRT-compatible, and casting an elided-lifetime `extern "C" fn(&AnyObject, ...)` produces a `for<'a>` HRT type that fails the bound. Raw pointers carry no lifetime, dodge the issue, and faithfully match the `objc_msgSend` ABI.
- Callbacks marked `extern "C-unwind"` rather than `extern "C"`. Plain `extern "C"` is undefined behavior if a foreign exception unwinds through it (Itanium ABI on macOS Apple Silicon makes this concrete). objc2's own examples use `extern "C-unwind"` throughout.
- `bool` is not `objc2::Encode` (objc2 docs: *"bool [...] cannot be safely used in custom defined methods"*), so BOOL-returning callbacks return `objc2::runtime::Bool` and use `.as_bool()` at comparison sites.
- Local `CgPoint` / `CgSize` / `CgRect` `#[repr(C)]` wrappers for the geometric types used as `msg_send!` returns. cocoa's `NSPoint` / `NSSize` / `NSRect` are external types and can't implement objc2's external `Encode` trait (orphan rule), so the wrappers carry the impl. `From` conversions to/from cocoa types at the boundary.
- `ClassBuilder::new`, `class.add_ivar`, and `AnyClass::get` / `AnyProtocol::get` take `&CStr` (was `&str` in objc 0.2). `CString::new(...)` for the runtime-built per-instance class name; `CStr::from_bytes_with_nul(b"...\0").unwrap()` for compile-time-known names (avoids `c"..."` literals which need edition 2021+).

## Latent bugs surfaced and fixed

The migration's stricter unwinding (`extern "C-unwind"`) and stricter `@encode` verification (`debug_assertions` mode) surfaced several bugs that have been latent in `master` for a while. I'm fixing them inside this PR rather than as separate PRs because they're directly entangled with the migration's stricter checks.

1. **`-[NSEvent isARepeat]` called eagerly on `flagsChanged:` events** (`keyboard.rs::process_native_event`). Per [Apple docs](https://developer.apple.com/documentation/appkit/nsevent/1525068-isarepeat), `isARepeat` raises `NSInternalInconsistencyException` when sent to non-key events. The result was only used for `event_type == NSKeyDown` (`&&` short-circuit), but the call was unconditional. Under `extern "C"` + objc 0.2 the foreign exception's behavior on macOS aarch64 happened to be survivable; under `extern "C-unwind"` the unwind is well-defined and silently swallows the modifier-state KeyboardEvent for every Alt/Cmd/Shift press, breaking downstream modifier tracking (textbox consumers can't see Cmd+Backspace as `cx.modifiers.logo()`). Verified end-to-end: an instrumented trace showed `flagsChanged ENTER` firing then no follow-up; gating on `event_type == NSKeyDown` restores the modifier-press KeyboardEvent.

2. **`-[NSEvent keyCode]` called eagerly before event-type filter** (`keyboard.rs::process_native_event`). Same exception class as `isARepeat` ([Apple docs](https://developer.apple.com/documentation/appkit/nsevent/1534513-keycode): *"raises an `NSInternalInconsistencyException` if sent to non-key events."*) The function already had a `_ => return None` arm acknowledging that AppKit occasionally dispatches non-key events into `keyDown:`/`keyUp:`/`flagsChanged:` selectors (`NSAppKitDefined`, `NSSystemDefined`, sync events around Cmd-Tab and input-source switches), but `keyCode()` was called before reaching that match, making the fallback dead code. Fix: hoist event-type filter above `keyCode()`.

3. **`NSArray::objectAtIndex(tracking_areas, 0)` without count check** (`view.rs::update_tracking_areas`). Raises `NSRangeException` if `index >= count`. The companion site in `view_will_move_to_window` already guards on `count != 0`; mirror the same guard here for defense-in-depth.

4. **Three return-type vs Apple `@encode` mismatches** surfaced by objc2's debug-mode runtime verification:
   - `view.rs::create_view`: `registerForDraggedTypes:` declared `@` (id), Apple's signature returns `void` (`v`). Fix: declare `()`.
   - `window.rs::open_parented`: `addSubview:` declared `@`, Apple returns `void`. Fix: declare `()`.
   - `gl/macos.rs::create`: `retain` declared `()`, Apple returns `instancetype` (`@`). Fix: declare `*mut AnyObject`.

   objc 0.2's `msg_send!` does not verify return-type encodings; objc2's does (under `debug_assertions`), so these mismatches that have shipped silently now panic on first debug run of `cargo run --example open_window` and `cargo run --example render_femtovg`.

## Test plan

- [x] `cargo build --workspace --all-targets --all-features` with `RUSTFLAGS=-D warnings` clean
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo test --workspace --all-targets --all-features` clean
- [x] `cargo run --example open_window` opens cleanly, dispatches mouse + keyboard events, modifier-state tracking works (Alt+G correctly synthesizes `©` with `Modifiers(ALT)`, Alt release synthesizes `Modifiers(0x0)`)
- [x] `cargo run --example render_femtovg` opens cleanly, OpenGL rendering through `gl/macos.rs` works, mouse events dispatch
- [x] DAW integration test against an in-tree consumer using `Window::open_parented` with Skia/Metal: plugin opens cleanly in Ableton Live + REAPER (with and without "Send all keyboard input to plug-in"); all keyboard/mouse/drag-drop functionality verified including modifier-held shortcuts (Cmd+Backspace, Alt+Backspace, Cmd+Left, etc.); `pluginval --strictness-level 10` passes
- [ ] `cargo run --example open_parented` panics in `softbuffer/cg.rs:36:26 null pointer dereference`. Confirmed unchanged from `master` (same panic on a clean checkout of the pre-migration code), so this is a pre-existing softbuffer-vs-baseview lifecycle issue, not a migration regression. Worth a separate issue/PR.

## What's not in this PR

Intentionally minimal. cocoa is kept as the source of framework type bindings; `cocoa::base::id` is kept at the public surface for raw pointer crossings; `Retained<T>` and `MainThreadMarker` patterns are NOT introduced. A follow-up PR can replace cocoa with `objc2-app-kit` / `objc2-foundation` for typed framework methods, modern memory model, and current-Xcode SDK bindings, but that's a larger refactor with its own design discussion (and probably the right call to align on with @micahrj before pursuing).

Related to #229. Supersedes #231.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ uuid = { version = "0.8", features = ["v4"], optional = true }
 [target.'cfg(target_os="macos")'.dependencies]
 cocoa = "0.24.0"
 core-foundation = "0.9.1"
-objc = "0.2.7"
+objc2 = "0.6"
 uuid = { version = "0.8", features = ["v4"] }
 
 [dev-dependencies]
--- a/src/gl/macos.rs
+++ b/src/gl/macos.rs
@@ -1,7 +1,3 @@
-// This is required because the objc crate is causing a lot of warnings: https://github.com/SSheldon/rust-objc/issues/125
-// Eventually we should migrate to the objc2 crate and remove this.
-#![allow(unexpected_cfgs)]
-
 use std::ffi::c_void;
 use std::str::FromStr;
 
@@ -21,7 +17,8 @@ use core_foundation::base::TCFType;
 use core_foundation::bundle::{CFBundleGetBundleWithIdentifier, CFBundleGetFunctionPointerForName};
 use core_foundation::string::CFString;
 
-use objc::{msg_send, sel, sel_impl};
+use objc2::msg_send;
+use objc2::runtime::AnyObject;
 
 use super::{GlConfig, GlError, Profile};
 
@@ -98,15 +95,16 @@ impl GlContext {
         NSOpenGLView::display_(view);
         parent_view.addSubview_(view);
 
-        let context: id = msg_send![view, openGLContext];
-        let () = msg_send![context, retain];
+        let context_any: *mut AnyObject = msg_send![view as *mut AnyObject, openGLContext];
+        let _: *mut AnyObject = msg_send![context_any, retain];
+        let context: id = context_any as id;
 
         context.setValues_forParameter_(
             &(config.vsync as i32),
             NSOpenGLContextParameter::NSOpenGLCPSwapInterval,
         );
 
-        let () = msg_send![pixel_format, release];
+        let () = msg_send![pixel_format as *mut AnyObject, release];
 
         Ok(GlContext { view, context })
     }
@@ -131,7 +129,7 @@ impl GlContext {
     pub fn swap_buffers(&self) {
         unsafe {
             self.context.flushBuffer();
-            let () = msg_send![self.view, setNeedsDisplay: YES];
+            let () = msg_send![self.view as *mut AnyObject, setNeedsDisplay: YES];
         }
     }
 
@@ -139,7 +137,7 @@ impl GlContext {
     pub(crate) fn resize(&self, size: NSSize) {
         unsafe { NSView::setFrameSize(self.view, size) };
         unsafe {
-            let _: () = msg_send![self.view, setNeedsDisplay: YES];
+            let _: () = msg_send![self.view as *mut AnyObject, setNeedsDisplay: YES];
         }
     }
 }
@@ -147,8 +145,8 @@ impl GlContext {
 impl Drop for GlContext {
     fn drop(&mut self) {
         unsafe {
-            let () = msg_send![self.context, release];
-            let () = msg_send![self.view, release];
+            let () = msg_send![self.context as *mut AnyObject, release];
+            let () = msg_send![self.view as *mut AnyObject, release];
         }
     }
 }
--- a/src/macos/keyboard.rs
+++ b/src/macos/keyboard.rs
@@ -24,7 +24,8 @@ use cocoa::appkit::{NSEvent, NSEventModifierFlags, NSEventType};
 use cocoa::base::id;
 use cocoa::foundation::NSString;
 use keyboard_types::{Code, Key, KeyState, KeyboardEvent, Modifiers};
-use objc::{msg_send, sel, sel_impl};
+use objc2::msg_send;
+use objc2::runtime::AnyObject;
 
 use crate::keyboard::code_to_location;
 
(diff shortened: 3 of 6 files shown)


micahrj:
Forgive me if my assumption is incorrect, but this looks like a largely LLM-generated PR. This is much too big and consequential a change for me to trust an LLM to do it, and properly reviewing it will take as much work or more than just making the change myself. I would also rather make a complete move from `objc` and `cocoa` to the `objc2` ecosystem rather than doing it piecemeal like this. Since this change is not actually necessary to fix the failing CI builds (#233 is enough), I am going to close this PR.

PR #238, opened 2026-05-03
macOS: Set ns_window to the parent window when creating WindowHandles

prokopyl:
Currently, when baseview creates a window in parented mode, the `WindowHandle`s it gives out have their `ns_window` field set to `null`, which makes some downstream libraries (such as `softbuffer`) crash.

This means the `open_parented` example completely crashes as a result, it does not even start properly on macOS.

This PR fixes this by storing the parent `NSWindow` pointer separately, and then using it to set the `WindowHandle`'s `ns_window` field.

I can confirm it prevents the crash on macOS 26.2.

This PR supersedes the fix in #219. Unlike in that PR, this PR stores the parent `NSWindow` pointer separately, so it does not impact other baseview operations.

### DIFF
--- a/src/macos/window.rs
+++ b/src/macos/window.rs
@@ -113,9 +113,9 @@ impl WindowInner {
 
     fn raw_window_handle(&self) -> RawWindowHandle {
         if self.open.get() {
-            let ns_window = self.ns_window.get()
-                .or(self.parent_ns_window)
-                .unwrap_or(ptr::null_mut()) as *mut c_void;
+            let ns_window =
+                self.ns_window.get().or(self.parent_ns_window).unwrap_or(ptr::null_mut())
+                    as *mut c_void;
 
             let mut handle = AppKitWindowHandle::empty();
             handle.ns_window = ns_window;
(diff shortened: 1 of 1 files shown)


dathinaios:
Tested both examples and can confirm it is not crashing anymore 👍

Mac Studio M1 Max
MacOS: 15.7.3 (24G419)

prokopyl:
@dathinaios Sweet! Thank you for taking the time to test it ^^

micahrj:
I don't think this is the right thing to do. If a downstream library is going to do nontrivial things with the `NSWindow` in the `WindowHandle`, we should not be passing it the parent window, since that parent window belongs to the host and we should avoid doing arbitrary things with it. If a downstream library such as `softbuffer` does not actually need a non-null `NSWindow` to function properly, then the issue should be addressed in that library rather than in `baseview`.

micahrj:
> This means the `open_parented` example completely crashes as a result, it does not even start properly on macOS.

I think this should be addressed by updating to the latest version of `softbuffer`, where the issue should not occur.

PR #240, opened 2026-05-04
macOS: move from `objc` and `cocoa` crates to the `objc2` crates

prokopyl:
This PR removes the `cocoa`, `core_foundation` and `objc` crates as dependencies, and replaces them with the `objc2` family of crates.

On top of removing outdated dependencies and bringing us closer to the rest of the ecosystem, `objc2` brings a couple benefits:
* Proper types throughout the codebase. `objc` exposed a *lot* of things as just `id`, and didn't manage `null` nor proper object retention at all.
* Way fewer instances of `unsafe`. `objc2` provides simple safe wrappers for a lot of the simple operations.

The goal of this PR is to update the invocations while changing as little logic as possible. Further refactorings and code cleanups are needed, but will come in later releases.

I have tested those changes successfully on macOS 26.2. More testing is always welcome however. :slightly_smiling_face:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,13 @@ exclude = [".github"]
 
 [features]
 default = []
-opengl = ["uuid"]
+opengl = [
+    "uuid",
+    "objc2-core-foundation/CFBundle",
+    "objc2-app-kit/NSOpenGL",
+    "objc2-app-kit/NSOpenGLView",
+    "objc2-app-kit/objc2-open-gl"
+]
 
 [dependencies]
 keyboard-types = { version = "0.6.1", default-features = false }
@@ -41,7 +47,7 @@ uuid = { version = "0.8", features = ["v4"], optional = true }
 [target.'cfg(target_os="macos")'.dependencies]
 uuid = { version = "1.23.1", features = ["v4"] }
 objc2 = "0.6.4"
-objc2-core-foundation = { version = "0.3.2", default-features = false, features = ["std"] }
+objc2-core-foundation = { version = "0.3.2", default-features = false, features = ["std", "CFString"] }
 objc2-foundation = { version = "0.3.2", default-features = false, features = ["std", "NSEnumerator"] }
 objc2-app-kit = { version = "0.3.2", default-features = false, features = [
     "NSApplication",
@@ -54,7 +60,6 @@ objc2-app-kit = { version = "0.3.2", default-features = false, features = [
     "NSTrackingArea",
     "NSView",
     "NSWindow",
-    "objc2-open-gl",
     "objc2-core-foundation"
 ] }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ uuid = { version = "0.8", features = ["v4"] }
 objc2 = "0.6.4"
 objc2-core-foundation = "0.3.2"
 objc2-foundation = "0.3.2"
-objc2-app-kit = "0.3.2"
+objc2-app-kit = { version = "0.3.2", features = ["objc2-open-gl"] }
 
 [dev-dependencies]
 rtrb = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ authors = [
     "Robbert van der Helm <mail@robbertvanderhelm.nl>",
     "Adrien Prokopowicz <prokopylmc@gmail.com>"
 ]
-edition = "2024"
+edition = "2021"
 license = "MIT OR Apache-2.0"
 description = "Low-level windowing system geared towards making audio plugin UIs."
 keywords = ["windowing", "audio", "plugin"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,12 +39,24 @@ winapi = { version = "0.3.8", features = ["libloaderapi", "winuser", "windef", "
 uuid = { version = "0.8", features = ["v4"], optional = true }
 
 [target.'cfg(target_os="macos")'.dependencies]
-uuid = { version = "0.8", features = ["v4"] }
-# TODO: only enable needed features
+uuid = { version = "1.23.1", features = ["v4"] }
 objc2 = "0.6.4"
-objc2-core-foundation = "0.3.2"
-objc2-foundation = "0.3.2"
-objc2-app-kit = { version = "0.3.2", features = ["objc2-open-gl"] }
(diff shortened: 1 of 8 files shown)


src/macos/window.rs
```diff
@@ -191,75 +200,72 @@ impl<'a> Window<'a> {
...
 
         let _ = Self::init(window_inner, window_info, build);
 
-        unsafe {
-            ns_window.setContentView_(ns_view);
-            ns_window.setDelegate_(ns_view);
-
-            let () = msg_send![pool, drain];
+        ns_window.setContentView(Some(&ns_view));
+        // TODO: this was here before, but couldn't have worked??
```
micahrj: Why not?

src/macos/window.rs
```diff
@@ -271,21 +277,27 @@ impl<'a> Window<'a> {
...
+            frame_timer: RetainedCell::empty(),
             window_info: Cell::new(window_info),
             deferred_events: RefCell::default(),
         });
 
         let window_state_ptr = Rc::into_raw(Rc::clone(&window_state));
 
         unsafe {
-            (*ns_view).set_ivar(BASEVIEW_STATE_IVAR, window_state_ptr as *const c_void);
+            // TODO: Pretty certain this is a cyclic reference (aaaa)
```
micahrj: It is a cyclic reference. It gets broken in `WindowInner::close`: https://github.com/RustAudio/baseview/blob/4cabe2f1595402072e196a1f4cf67fb324f94573/src/macos/window.rs#L77

src/macos/view.rs
```diff
@@ -93,258 +79,258 @@ macro_rules! add_simple_keyboard_class_method {
...
+    let class = this.class();
 
-        // Delete class
-        ::objc::runtime::objc_disposeClassPair(class);
+    if let Some(superclass) = class.superclass() {
+        let () = unsafe { msg_send![super(this, superclass), dealloc] };
     }
+
+    // Delete class
+    // SAFETY: TODO: nope, this is NOT sound, as this invalidates any &AnyClass
```
micahrj: Discussed here: https://github.com/madsmtm/objc2/issues/502

I think it's fine as long as Baseview keeps the class private.

Cargo.toml
```diff
@@ -12,7 +12,7 @@ authors = [
     "Robbert van der Helm <mail@robbertvanderhelm.nl>",
     "Adrien Prokopowicz <prokopylmc@gmail.com>"
 ]
-edition = "2018"
+edition = "2021"
```
micahrj, line 15: Any reason for doing this in the same change as migrating to `objc2`?

src/gl/macos.rs
```diff
@@ -112,47 +103,35 @@ impl GlContext {
     }
 
     pub unsafe fn make_not_current(&self) {
-        NSOpenGLContext::clearCurrentContext(self.context);
+        if Some(&self.context) == NSOpenGLContext::currentContext().as_ref() {
```
micahrj: Since this changes the logic from how it was before, just wanted to confirm that you've specifically tested that this still works (that the equality check does what you'd expect).

Cargo.toml
prokopyl: This was to enable using C-string literals, but I can move it to another PR if you'd prefer

src/gl/macos.rs
prokopyl: I did test it, but I actually didn't intend to change the logic from how it was before.

I was under the impression that `objc`'s `clearCurrentContext` was only clearing the given context, for some reason. But double checking I realized the given context was just unused, so I reverted that change

Cargo.toml
micahrj: No need, was just curious what it was for

src/macos/window.rs
```diff
@@ -140,8 +155,6 @@ impl<'a> Window<'a> {
         B: FnOnce(&mut crate::Window) -> H,
         B: Send + 'static,
     {
-        let pool = unsafe { NSAutoreleasePool::new(nil) };
```
micahrj, line 143: Why have all the `NSAutoreleasePool`s been removed?

src/macos/view.rs
```diff
@@ -56,28 +47,23 @@ macro_rules! add_simple_mouse_class_method {
...
-                modifiers: make_modifiers(modifiers),
+                modifiers: make_modifiers(event.modifierFlags()),
             }));
         }
 
-        $class.add_method(
-            sel!($sel:),
-            $sel as extern "C" fn(&Object, Sel, id),
-        );
+        $class.add_method(sel!($sel:),$sel as extern "C-unwind" fn(_, _, _) -> _);
```
micahrj:
```suggestion
        $class.add_method(sel!($sel:), $sel as extern "C-unwind" fn(_, _, _) -> _);
```

src/macos/view.rs
```diff
@@ -93,258 +79,271 @@ macro_rules! add_simple_keyboard_class_method {
             }
         }
 
-        $class.add_method(
-            sel!($sel:),
-            $sel as extern "C" fn(&Object, Sel, id),
-        );
+        $class.add_method(sel!($sel:),$sel as extern "C-unwind" fn(_, _, _) -> _);
```
micahrj:
```suggestion
        $class.add_method(sel!($sel:), $sel as extern "C-unwind" fn(_, _, _) -> _);
```

src/macos/window.rs
prokopyl: My bad, I didn't realize the way these worked (I assumed the `pool` variable being unused meant the pool wasn't necessary). I fixed this by replacing it with objc2's `autoreleasepool` calls (which are more resilient to errors/panics anyway)

src/macos/view.rs
```diff
@@ -427,95 +419,94 @@ extern "C" fn view_will_move_to_window(this: &Object, _self: Sel, new_window: id
...
+    };
 
-        let mut files = vec![];
-        for i in 0..NSArray::count(file_list) {
-            let data = NSArray::objectAtIndex(file_list, i);
-            files.push(from_nsstring(data).into());
-        }
+    let files = file_list
+        .into_iter()
+        .filter_map(|i| i.downcast::<NSString>().ok())
```
micahrj:
```suggestion
        .filter_map(|s| s.downcast::<NSString>().ok())
```
Or something else other than `i`, since it is not an index.

micahrj (approved):

PR #249, opened 2026-05-23
Windows: Fix DND on non-parented windows (bugfix)

AlexCharlton:
Hi Rustaudio, I'm opening up a slew of PRs so I figured I'd provide a small preamble for this first one. I had forked baseview years ago and quickly diverged enough that I never wound up trying to submit anything upstream. With the 0.1 release I finally decided get on board and contribute back.

The majority of these are relatively small bugfixes for Windows, though a few are new features – The PR descriptions will indicate what the contents contain.

Thanks in advance for the reviews!

***

Filtering GetMessageW to hwnd precludes getting events from the OLE thread that's handling the DND. In practice this means that when dragging something into the baseview window, the drag process (that OLE thread) freezes up, making it impossible to complete the drag.

### DIFF
--- a/src/win/window.rs
+++ b/src/win/window.rs
@@ -512,7 +512,7 @@ pub(super) struct WindowState {
     window_class: Cell<Option<RegisteredClass>>,
     current_size: Cell<PhySize>,
     current_scale_factor: Cell<f64>,
-    _parent_handle: Option<ParentHandle>,
+    _parent_handle: ParentHandle,
     keyboard_state: RefCell<KeyboardState>,
     mouse_button_counter: Cell<usize>,
     mouse_was_outside_window: RefCell<bool>,
@@ -632,7 +632,7 @@ impl Window<'_> {
         B: FnOnce(&mut crate::Window) -> H,
         B: Send + 'static,
     {
-        let (_, _) = Self::open(false, null_mut(), options, build);
+        let (window_handle, _) = Self::open(false, null_mut(), options, build);
 
         unsafe {
             let mut msg: MSG = std::mem::zeroed();
@@ -646,6 +646,10 @@ impl Window<'_> {
 
                 TranslateMessage(&msg);
                 DispatchMessageW(&msg);
+
+                if !window_handle.is_open() {
+                    break;
+                }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


prokopyl:
Hi! And thanks for submitting your changes, it is much appreciated! 

Your PRs have quite a variety of changes between bug fixes, new features, platform support, some breaking changes and other improvements, so I'm going to order them according to our [short-term roadmap](https://github.com/RustAudio/baseview/milestones), and then review them in that priority order.

That means I'm likely gonna make a few minor releases before I'll get to your bigger PRs. :slightly_smiling_face:

src/win/window.rs
```diff
@@ -603,13 +603,13 @@ impl Window<'_> {
...
     {
-        let (_, hwnd) = Self::open(false, null_mut(), options, build);
+        let (_, _) = Self::open(false, null_mut(), options, build);
 
         unsafe {
             let mut msg: MSG = std::mem::zeroed();
 
             loop {
-                let status = GetMessageW(&mut msg, hwnd, 0, 0);
+                let status = GetMessageW(&mut msg, null_mut(), 0, 0);
```
prokopyl, line 641: I agree that removing the filtering here is correct, however in this case that means there is no other condition to break from the loop.
So when the window is closed this loops keep running and the program doesn't terminate correctly.

You should add another condition in the loop (after `DispatchMessage`) to check the `is_open` boolean (from the `ParentHandle` returned by `open` above), and break if that window isn't open anymore.

prokopyl (changes requested):

AlexCharlton:
> Your PRs have quite a variety of changes between bug fixes, new features, platform support, some breaking changes and other improvements, so I'm going to order them according to our [short-term roadmap](https://github.com/RustAudio/baseview/milestones), and then review them in that priority order.

Thanks for reviewing, @prokopyl! This makes sense to me!

src/win/window.rs
AlexCharlton: I clearly did notice this at some point, which is why I had added https://github.com/RustAudio/baseview/pull/250, but I evidently didn't realize they were related! 🤦

Good catch!

AlexCharlton: A consequence of this change is that we need to keep the `parent_handle` alive, or else it gets dropped immediately and thus `is_open` will be false. Doing so meant `WindowState._parent_handle` is never `None`, so I removed the `Option` wrapping it.

prokopyl (approved):
Tested and looks good to me now! Thank you!

PR #250, opened 2026-05-23
Windows: Don't call GetMessageW after window has closed (bugfix)

AlexCharlton:
This fixes an occasional panic: sometimes we'll re-enter the message-handling loop after a window has been closed, and GetMessageW will fail.


AlexCharlton:
Issue caused by https://github.com/RustAudio/baseview/pull/249 
Will fix there

PR #253, opened 2026-05-23
Windows: Fix drag mouse position for parented windows (bugfix)

AlexCharlton:
The drag position was being reported corrected on unparented windows, but it could be completely off for parented windows, depending on the parent. Rather than relying on the cursor position as reported by OLE, this calls `GetCursorPos` instead.

### DIFF
--- a/src/platform/win/drop_target.rs
+++ b/src/platform/win/drop_target.rs
@@ -10,7 +10,12 @@ use windows::Win32::System::Ole::*;
 use windows::Win32::System::SystemServices::MODIFIERKEYS_FLAGS;
 use windows_core::Ref;
 use windows_sys::Win32::{
-    Foundation::POINT, Graphics::Gdi::ScreenToClient, UI::Shell::DragQueryFileW,
+    Foundation::POINT,
+    Graphics::Gdi::ScreenToClient,
+    UI::{
+        HiDpi::{LogicalToPhysicalPointForPerMonitorDPI, PhysicalToLogicalPointForPerMonitorDPI},
+        Shell::DragQueryFileW,
+    },
 };
 
 use crate::{DropData, DropEffect, Event, EventStatus, MouseEvent, PhyPoint, Point};
@@ -63,7 +68,19 @@ impl DropTarget {
             return;
         };
         let mut pt = POINT { x: pt.x, y: pt.y };
-        unsafe { ScreenToClient(window_state.hwnd, &mut pt as *mut POINT) };
+
+        // ScreenToClient isn't DPI-aware
+        if unsafe { PhysicalToLogicalPointForPerMonitorDPI(null_mut(), &mut pt) == 0 } {
+            return;
+        }
+
+        if unsafe { ScreenToClient(window_state.hwnd, &mut pt) == 0 } {
+            return;
+        }
+
+        if unsafe { LogicalToPhysicalPointForPerMonitorDPI(null_mut(), &mut pt) == 0 } {
+            return;
+        }
         let phy_point = PhyPoint::new(pt.x, pt.y);
         self.drag_position.set(phy_point.to_logical(&window_state.window_info()));
     }
--- a/src/platform/win/drop_target.rs
+++ b/src/platform/win/drop_target.rs
@@ -10,8 +10,10 @@ use windows::Win32::System::Ole::*;
 use windows::Win32::System::SystemServices::MODIFIERKEYS_FLAGS;
 use windows_core::Ref;
 use windows_sys::Win32::{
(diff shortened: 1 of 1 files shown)


AlexCharlton:
Waiting for https://github.com/RustAudio/baseview/pull/249, which will conflict

prokopyl:
Your comments in the code mention having to work around DPI virtualization, and that seems quite weird to me since our windows should be DPI-aware.

Can you provide an example which reproduces this issue? (I'm guessing it's probably with a specific host that isn't DPI-aware for some reason?)

AlexCharlton:
I ran into this problem on Bitwig, which launches plugins from one or more isolated processes. I'm surmising that _these_ processes aren't DPI aware and it's this parent that the OLE thread is attached to and which generates the drag points, hence the coordinate mismatch. This is conjecture, but it would explain why the passed-in points are not in the same reference frame as the points fetched by `GetCursorPos`.

Your comment made me realize that I really ought to test this on a host with a different way of handling plugins, and lo and behold, this solution does not hold water with Reaper. So, back to the drawing board. I'll keep this in draft until I figure out a solution that works more broadly.

AlexCharlton:
Getting this working was more convoluted than I imagined, but I'll describe what I encountered, and how I wound up here.

There's two different issues being fixed. 

First is the coordinates of the point being given by the OLE process. For non-parented windows, everything seemed kosher. For windows with parents, the results differ by host. In Reaper, the OLE point was as expected: After `ScreenToClient`, they have the correct reference frame and are in physical pixels. In Bitwig, they were shifted to some frame of reference that I can only guess at. For instance, at the top-left of my window I was getting 490x-383 after converting ScreenToClient. Using `GetCursorPos` to get the point addressed this issue.

Second, the scale of the point returned by `GetCursorPos` differs based on the combination of the parent AND the drag source. With Reaper as the parent, the `GetCursorPos` was reliably in physical pixels no matter the drag source. In Bitwig, a custom (baseview) drag source was resulting in physical pixels, while using Windows Explorer as a drag source resulted in _logical_ pixels.

This made it necessary to detect whether the specific drag context would return physical or logical pixels, which I accomplished by calling `GetClientRect` and comparing the results to the dimensions held by the window state. `GetClientRect` returned a rect scaled in the same manor as `GetCursorPos`, so it either returns physical or logical pixels depending on the drag context. This result of this detection is cached on the `DropTarget`, so it only happens once per drag.

Why this is so damn hairy is beyond me, and if you have any other ideas, I'm all ears.

prokopyl:
I agree this seems *way* too hairy, and while I can reproduce your findings, I cannot find any documentation that supports them.

I suspect this may be related to `baseview`'s current DPI handling. Right now it uses `SetProcessDpiAwarenessContext` and lets it silently fail, meaning it will fail to override hosts that have some kind of DPI awareness. The current implementation also fails to take DPI into account in a variety of places.

I've kept this behavior for now because those will likely be breaking changes, so I've slated them for `0.2`. I think I'm going to put a pin in that bug too, until I fix the DPI handling, and maybe that will help make this make more sense.

AlexCharlton:
Confirmed that this solution works with FL Studio, and Ableton hosts too.

prokopyl:
I just finished working on #266 for proper DPI awareness, can you check if that itself fixes the issue? Or perhaps it at least makes the situation less hairy hopefully.

AlexCharlton:
I'm afraid #266 hasn't fixed this issue: both underlying problems remain. This PR continues to address the issues.

I don't suppose your recent work has given you any insight into a cleaner way of accomplishing this?

prokopyl:
I've been going through ilmai's [`plugin-canvas`](https://github.com/ilmai/plugin-things/), and it seems they stumbled onto a similar issue.

According to [this implementation](https://github.com/ilmai/plugin-things/blob/daf6cbda7e82e073bfa8cf6382d09b8f6cb21a45/plugin-canvas/src/platform/win32/drop_target.rs#L93), it seems that `ScreenToClient` isn't DPI-aware and that may be the source of those problems. They use `PhysicalToLogicalPointForPerMonitorDPI`/`LogicalToPhysicalPointForPerMonitorDPI` around it to work around this issue.

Perhaps that solution could work for us as well?

AlexCharlton:
That does the trick perfectly. Great find, and thank you ilmai for the solution!

Just a heads up that I tested this fix in Bitwig, Reaper, FL Studio, and Cubase, but my test plugin crashed Ableton. Reverting to pre #266 I could get the plugin to load, so now I'm bisecting to see what exactly is causing Ableton to be sad. I'll open a bug when I figure it out. I'm noting this here since testing this on Ableton on a post-DPI-aware branch remains TODO.

AlexCharlton:
Opened https://github.com/RustAudio/baseview/issues/274

PR #256, opened 2026-05-23
Add `WindowOpenOptions.resizable` (feature)

AlexCharlton:
This adds a new `WindowOpenOptions`, allowing the user to specify when a window should not be resizable (on all platforms). The default behaviour (`resizable: true`) is unchanged.

### DIFF
--- a/src/macos/window.rs
+++ b/src/macos/window.rs
@@ -224,14 +224,20 @@ impl<'a> Window<'a> {
 
             let window_info = WindowInfo::from_logical_size(options.size, scaling);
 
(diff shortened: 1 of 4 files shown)


prokopyl:
I tried to rebase your PR but there have been a lot of internal changes since then, and a Git merge just results in an empty commit. :slightly_frowning_face: 

I had to re-implement the logic in #310, but thank you for your PR still, it was quite useful!

AlexCharlton:
Thanks for re-implementing! 🙏

PR #259, opened 2026-05-24
Win32: Introduce safe wrapper for window creation process and lifecycle

prokopyl:
This PR introduces an internal safe wrapper (`WindowImpl`) around the Win32 window creation process and lifecycle.

This also makes it possible to properly handle messages sent to the Baseview window while processing the `WM_CREATE` and `WM_DESTROY` messages. It also makes it possible to safely send, receive and process `WM_DESTROY` messages during any event handler.

This means the Baseview implementation (or user implementation) can now perform non-trivial work during window creation, and any generated messages (e.g. window resizing) will not be missed by Baseview's internal handlers.

This PR also moves the initialization process from after `CreateWindow` is called to during the processing of the `WM_CREATE` message.


src/wrappers/win32/window/proc.rs
```diff
@@ -0,0 +1,106 @@
...
+        _ => {
+            let Some(inner_ptr) = window.get_userdata_ptr::<WindowData<W>>() else {
+                return handle_default();
+            };
+
+            // This guarantees WindowData remains valid until the end of this scope,
+            // even if the event handler leads to the window being destroyed
+            let inner = unsafe { WindowData::from_raw(inner_ptr) };
+
+            let result = catch_unwind(AssertUnwindSafe(|| {
```
micahrj: What is the purpose of this `catch_unwind` (and all the other uses in `wnd_proc`? As of [Rust 1.81](https://blog.rust-lang.org/2024/09/05/Rust-1.81.0/#abort-on-uncaught-panics-in-extern-c-functions), it is sound to panic inside `extern "C"` functions, as the uncaught panics will just get turned into an abort. It kind of seems more sensible to me to just default to an abort than to swallow panics and close the window.

PR #260, opened 2026-05-25
Windows: Pass `EventStatus::Ignored` keyboard events to parent (feature parity with OSX)

AlexCharlton:
Title says it all, I think :)

### DIFF
--- a/src/win/window.rs
+++ b/src/win/window.rs
@@ -13,15 +13,16 @@ use windows_sys::Win32::{
         },
         WindowsAndMessaging::{
             AdjustWindowRectEx, CreateWindowExW, DefWindowProcW, DestroyWindow, DispatchMessageW,
-            GetMessageW, GetWindowLongPtrW, LoadCursorW, PostMessageW, SetCursor, SetTimer,
-            SetWindowLongPtrW, SetWindowPos, TranslateMessage, GWLP_USERDATA, HTCLIENT, MSG,
-            SWP_NOACTIVATE, SWP_NOMOVE, SWP_NOZORDER, WHEEL_DELTA, WM_CHAR, WM_CLOSE, WM_CREATE,
-            WM_DPICHANGED, WM_INPUTLANGCHANGE, WM_KEYDOWN, WM_KEYUP, WM_KILLFOCUS, WM_LBUTTONDOWN,
-            WM_LBUTTONUP, WM_MBUTTONDOWN, WM_MBUTTONUP, WM_MOUSEHWHEEL, WM_MOUSEMOVE,
-            WM_MOUSEWHEEL, WM_NCDESTROY, WM_RBUTTONDOWN, WM_RBUTTONUP, WM_SETCURSOR, WM_SETFOCUS,
-            WM_SHOWWINDOW, WM_SIZE, WM_SYSCHAR, WM_SYSKEYDOWN, WM_SYSKEYUP, WM_TIMER, WM_USER,
-            WM_XBUTTONDOWN, WM_XBUTTONUP, WS_CAPTION, WS_CHILD, WS_CLIPSIBLINGS, WS_MAXIMIZEBOX,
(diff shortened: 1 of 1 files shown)


micahrj:
This is not going to work correctly when there is not a one-to-one relationship between window messages and baseview keyboard events. The `process_message` call immediately above these changes can collect up multiple window messages which correspond to a single logical keyboard event (e.g. in the case of dead keys in some keyboard layouts). When this happens, the effect of this PR will be that all but the final window message will always be forwarded to the parent, and only the final message will be optionally forwarded or not forwarded depending on the returned `EventStatus`.

AlexCharlton:
> This is not going to work correctly when there is not a one-to-one relationship between window messages and baseview keyboard events. The process_message call immediately above these changes can collect up multiple window messages which correspond to a single logical keyboard event (e.g. in the case of dead keys in some keyboard layouts). When this happens, the effect of this PR will be that all but the final window message will always be forwarded to the parent, and only the final message will be optionally forwarded or not forwarded depending on the returned EventStatus.

Ah ha! I did not appreciate that `process_message` was doing this. I suppose a solution here could be to stash the set of `(msg, wparam, lparam)` used to generate an event in `KeyboardState`, return them with the event, and then `SendMessageW` the lot of them when the event is ignored. Sound workable?

micahrj:
> Ah ha! I did not appreciate that `process_message` was doing this. I suppose a solution here could be to stash the set of `(msg, wparam, lparam)` used to generate an event in `KeyboardState`, return them with the event, and then `SendMessageW` the lot of them when the event is ignored. Sound workable?

That sounds reasonable to me, though I would want to see some testing for the specific multi-message cases that `process_message` handles.

More generally, I would want to see some testing of this PR in different hosts to determine if and where it is actually necessary and if it actually has the desired effect.

AlexCharlton:
> More generally, I would want to see some testing of this PR in different hosts to determine if and where it is actually necessary and if it actually has the desired effect.

By "see some testing," I'm guessing you mean you'd like me to describe under what conditions I tried this out?

As for "if and where it is actually necessary", I'll refer you to the original commit+comment that introduced this feature: https://github.com/RustAudio/baseview/commit/72b6a4a2d1662a16d94ca8087fcfd466be210b90#diff-14da7c350de5742eaa248de110fefeae9b4b639c0358e26024d6040ccf174a41R100-R105

There's lots of cases where this could be useful, though. Many plugins don't have any keybindings at all, so you really want to ignore everything or else you'll block your DAW's keybindings while your plugin is active. Or you could only handle keys that do something in your plugin, so that anything that doesn't overlap will still go through to your DAW.

micahrj:
I definitely understand the rationale for the feature. I'm more interested in examples of specific hosts where this particular approach actually has the desired effect. Keyboard focus handling in hosts and plugins on Windows is a real mess, with hosts implementing all sorts of workarounds (using Win32 subclassing to intercept keyboard events received by the plugin's child window, filtering out keyboard events at the point of the event loop, etc.) and plugins implementing all sorts of workarounds in turn (creating invisible child windows to handle keyboard events, registering keyboard hooks, etc.). Because of that, it is not obvious to me just from looking at the code whether a change like this will actually accomplish what it's supposed to without causing other problems.

AlexCharlton:
Ah, I see. Yeah, you're definitely right to question this. I had originally tested this against Bitwig, and it worked as you'd hope. Having just tried it in Reaper, the forwarded events don't seem to do anything at all. Swapping to using `PostMessageW` got it working in Reaper, but not in Bitwig. Deferring and then `Send/PostMessage` doesn't seem to change this. I'll keep plugging away to see if I can find something that solves both cases. I'll also grab some more DAWs for testing - I didn't appreciate just how varied the behavior could be!

AlexCharlton:
Noting for myself:
- `SendMessage` works on Bitwig, Cubase
- `PostMessage` works on Ableton, Reaper, Cubase
- Neither work on FL Studio

PR #265, opened 2026-06-13
run cargo fmt

TristanCacqueray:
This change fixes the CI.


prokopyl:
It looks like this does the opposite. Maybe you have some weird rustfmt config locally?

TristanCacqueray:
Indeed :/ sorry for the noise!

PR #267, opened 2026-06-14
Embrace reentrancy

prokopyl:
This PR allows the `WindowHandler`'s methods to be called re-entrantly, which allows for natural platform behavior on Windows and macOS.

This is done by changing the `WindowHandler` methods' signatures to only take `&self` instead of `&mut self`. This makes it so baseview users need to use interior mutability in their types (`Cell`, `OnceCell`, `RefCell`, etc.).

While this increases complexity for users, this allows for more granularity in the borrows (rather than `baseview` attempting to prevent reentrancy, and trying to mutably borrow the whole `WindowHandler` for each event).

This also allows for downstream users to perform any kind of action in any kind of handler (e.g. resizing the window on click in a menu), without having to "defer" anything.

### DIFF
--- a/examples/open_parented/src/main.rs
+++ b/examples/open_parented/src/main.rs
@@ -2,13 +2,14 @@ use baseview::{
     Event, EventStatus, PhySize, Window, WindowEvent, WindowHandle, WindowHandler,
     WindowOpenOptions,
 };
+use std::cell::{Cell, RefCell};
 use std::num::NonZeroU32;
 
 struct ParentWindowHandler {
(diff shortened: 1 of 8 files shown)


aidan729:
@prokopyl Any reason you are waiting to merge this into master?

prokopyl:
> @prokopyl Any reason you are waiting to merge this into master?

This is a breaking change (and a rather major one at that), so it's slated for the next major release of `baseview` (v0.2).

It won't be merged into `master` until that release is ready, but I'll merge this and other major PRs in the `next` branch.

PR #280, opened 2026-06-24
Implement macOS cursor setting and hiding support

SpruceCloud:
As mentioned in https://github.com/RustAudio/baseview/pull/211/ , merging the mouse changes as a separate commit.

Took a bit longer, but now I am focused on finishing the changes.

### DIFF
--- a/src/platform/macos/view.rs
+++ b/src/platform/macos/view.rs
@@ -11,7 +11,7 @@ use crate::{
 use objc2::__framework_prelude::Retained;
 use objc2::rc::Weak;
 use objc2::runtime::{NSObjectProtocol, ProtocolObject};
-use objc2::{msg_send, AllocAnyThread};
+use objc2::{msg_send, AllocAnyThread, ClassType};
 use objc2_app_kit::{
     NSApplication, NSDragOperation, NSDraggingInfo, NSEvent, NSFilenamesPboardType, NSTrackingArea,
     NSTrackingAreaOptions, NSView, NSWindow,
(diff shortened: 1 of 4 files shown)


src/platform/macos/view.rs
```diff
@@ -230,6 +230,11 @@ impl BaseviewView {
 impl Drop for BaseviewView {
     fn drop(&mut self) {
         self.state.closed.set(true);
+        if self.state.cursor_hidden.get() {
+            unsafe {
+                let _: () = objc2::msg_send![objc2_app_kit::NSCursor::class(), unhide];
```
prokopyl: This should use NSCursor::unhide directly, instead of using a raw, unsafe msg_send call

src/platform/macos/window.rs
```diff
@@ -156,7 +156,21 @@ impl<'a> Window<'a> {
 
     pub fn set_mouse_cursor(&self, cursor: MouseCursor) {
         let native_cursor = Cursor::from(cursor);
-        self.view.addCursorRect_cursor(self.view.bounds(), &native_cursor.load());
+        unsafe {
+            if let Some(loaded_cursor) = native_cursor.load() {
+                if self.inner.state.cursor_hidden.get() {
+                    let _: () = objc2::msg_send![objc2_app_kit::NSCursor::class(), unhide];
```
prokopyl: Same as above, this should use the safe APIs instead of msg_send here

prokopyl (changes requested):
Hi, and thanks for this PR!

The main issue I have is that you should use the safe APIs provided by `objc2`, instead of using the error-prone and unsafe `msg_send!` macro.

Other than that, this looks pretty good to me!

prokopyl (approved):

PR #301, opened 2026-07-24
Add HandlerError::from_boxed

BillyDM:
I found there was no easy way to create a `HandlerError` from an already boxed error.

Until Rust gets negative traits, we can't have an `impl From<Box<dyn Error>> for HandlerError` implementation. So for now I just added a `from_boxed` method.

There were also a couple of places with unecessary `'static` labels (`Box` already implies that the contents are `'static`), so I removed them.


prokopyl:
Good points, thank you for the PR! :smile:

PR #308, opened 2026-07-30
Make Hand cursor show a pointing hand on macOS

exa04:
This PR fixes `MouseCursor::Hand` incorrectly showing an open hand on macOS by making that cursor a `NSCursor::pointerHand` instead of a `NSCursor::openHand`.

This makes the macOS implementation of that cursor consistent with Windows and Linux which both show a pointing hand (respectively `IDC_HAND` and `hand2`)

Relevant docs:
https://developer.apple.com/documentation/appkit/nscursor


prokopyl (approved):
Thank you for the fix!

PR #318, opened 2026-08-18
Remove extraneous dbg! expression

kitlith:
This fixes building for release


kitlith:
We can change around exactly how we're doing a check against the release profile in CI, but I feel like something similar should be implemented if you want to keep conditionally enabling the Debug derive, to be able to catch future mistakes like this in CI before a release.

prokopyl:
Thanks for the fix!

PR #336, opened 2026-09-10
MacOS: Prevent automatic rescaling on resize events

davemollen:
Currently when the window is resized on macOS the UI scales proportionally.
Changing the CFRunLoop from kCFRunLoopDefaultMode to kCFRunLoopCommonModes removes that automatic rescaling behavior. Giving the user control over the scaling.

I came across this in the nice-plug example plugins. The expected behavior for these plugins is to not rescale the UI on resize events. What is actually happening is the UI gets rescaled on resize events and when resizing is finished the scale snaps back to a factor of 1 (the original scale).

### DIFF
--- a/src/wrappers/appkit/timer.rs
+++ b/src/wrappers/appkit/timer.rs
@@ -1,7 +1,7 @@
 use block2::RcBlock;
 use objc2::rc::Weak;
 use objc2_core_foundation::{
-    kCFAllocatorDefault, kCFRunLoopDefaultMode, CFRetained, CFRunLoop, CFRunLoopTimer,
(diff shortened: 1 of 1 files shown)


davemollen:
@prokopyl Could you have a look?

davemollen:
Here is a screen recording with the old kCFRunLoopDefaultMode:
https://github.com/user-attachments/assets/4c3ab2c1-efa1-46c9-9198-b47e4ca59579

And this is a recording of using kCFRunLoopCommonModes:
https://github.com/user-attachments/assets/de90c9f6-2dd6-4aeb-90ab-382360485f06

